PR #380, opened 2015-01-15
Fix #379: OAuth2Client http_build_query() parameters were causing issues with Google

### LINKED ISSUE #379: Google authentication problem on the master branch (invalid_request, missing parameter) with Drupal 7
hybridauth v2.3.0 works fine with Drupal 7 and Google Authentication, but I was having a weird error on the master branch.

I traced it to this commit: https://github.com/hybridauth/hybridauth/commit/d69729fcbd73b5c79c2183e8a9ca1d3102beab36 from #296.

I think that:

    if($params) curl_setopt( $ch, CURLOPT_POSTFIELDS, http_build_query($params) );

Should be:

    if($params) curl_setopt( $ch, CURLOPT_POSTFIELDS, http_build_query($params, '', '&') );

This fixed the issue for me. Strangely, doing a print_r() on the resulting string didn't show a visible difference between the two strings, but doing a bin2hex() on it generated different strings. I also noticed that most http_build_query() calls in other providers seemed to be using those parameters as well.

This is a great library, many thanks :-)

mlutfy:


AdwinTrave:
Thanks for the fix!

Issue #389, opened 2015-02-04
POST request to Google API fails


Remo:
HybridAuth itself works perfectly fine in my case, was a breath to set up. The code looks like this:

``` php
$config = HybridAuthConfig::get();
$hybridauth = new Hybrid_Auth($config);
$google = $hybridauth->authenticate("Google");
```

I've added some scopes to the config:

``` php
"Google" => array(
    "enabled" => true,
    "keys" => array("id" => $google->getApiKey(), "secret" => $google->getSecret()),
    "scope" => "https://www.googleapis.com/auth/userinfo.profile https://www.googleapis.com/auth/userinfo.email https://www.googleapis.com/auth/calendar"
)
```

With this, I can easily access the events I have in my calendar, this works fine:

``` php
$url = sprintf('https://www.googleapis.com/calendar/v3/calendars/%s/events', rawurlencode('mail@domain.tld'));  
$result = $google->api()->get($url); 
print_r($result);
```

The POST method however never works:

``` php
$data = [
    'start' => [
        'dateTime' => '2015-02-14T13:00:00+01:00',
    ],
    'end' => [
        'dateTime' => '2015-02-14T13:30:00+01:00',
    ],
    'summary' => 'summary',
    'description' => 'description',
    'location' => 'location',
];
$url = sprintf('https://www.googleapis.com/calendar/v3/calendars/%s/events', rawurlencode('mail@domain.tld'));  
$result = $google->api()->post($url, ($data));
print_r($result);
```

This always prints

```
stdClass Object
(
    [error] => stdClass Object
        (
            [errors] => Array
                (
                    [0] => stdClass Object
                        (
                            [domain] => global
                            [reason] => authError
                            [message] => Invalid Credentials
                            [locationType] => header
                            [location] => Authorization
                        )

                )

            [code] => 401
            [message] => Invalid Credentials
        )

)
```
- I tried to pass the API key using ?key=mykey
- I tried to undo this change https://github.com/hybridauth/hybridauth/issues/296
- I tried to use `json_encode` instead of `http_build_query`

Does anyone have an idea what I'm doing wrong? Maybe an example somewhere?

Remo:
**In short**
* I don't see why we should enforce a query string
* We should send an authentication header
* There should be a way to specify the content-type of the request

**The long version**

[...]

PR #390, opened 2015-02-04
send authentication header

Remo:
as mentioned in https://github.com/hybridauth/hybridauth/issues/389 I'm having problems calling any Google API method using the POST method.

It's still not working, but I believe I'm past the authentication by adding a authentication header.

### DIFF
--- a/hybridauth/Hybrid/thirdparty/OAuth/OAuth2Client.php
+++ b/hybridauth/Hybrid/thirdparty/OAuth/OAuth2Client.php
@@ -208,7 +208,9 @@ class OAuth2Client
 		curl_setopt($ch, CURLOPT_CONNECTTIMEOUT , $this->curl_connect_time_out );
 		curl_setopt($ch, CURLOPT_SSL_VERIFYPEER , $this->curl_ssl_verifypeer );
 		curl_setopt($ch, CURLOPT_SSL_VERIFYHOST , $this->curl_ssl_verifyhost );
-		curl_setopt($ch, CURLOPT_HTTPHEADER     , $this->curl_header );
+
+		$header = array_merge($this->curl_header, ['Authorization: OAuth ' . $this->access_token]);		
+		curl_setopt($ch, CURLOPT_HTTPHEADER     , $header);
 
 		if($this->curl_proxy){
 			curl_setopt( $ch, CURLOPT_PROXY        , $this->curl_proxy);


AdwinTrave:
How much have you tested it?

Remo:
I did run quite some tests with the google API and it doesn't seem to cause any problems with the twitter login process, but obviously I haven't tested it with every provider there is.
Maybe make an option to enable it to be on the safe side?

AdwinTrave:
Yeah, that would be nice at least for one minor version, just to be sure.

Shade-:
This is also required by WordPress API (adding an extra header parameter) but you can add your custom header to the specific provider you're having issues as the majority of the others don't require this patch to work.

Remo:
@Shade- do you know what providers require that header? It's a bit time consuming to test everyone /-: So far it's google and wordpress.. Adding a parameter to the OAuth class would at least have the benefit that one could enable it for any provider. Would be nice if it was automatically enabled where required, just not sure how to do that..

Shade-:
$this->api->curl_header = array("Authentication: OAuth " . $this->access_token);

This would do the trick in your case (add this in the provider-specific file). But as I said before, some providers do require this and some others don't, so I'd say it's better to keep the existing configuration in the curl setup and add the extra header just to providers that need this patch. As far as I am concerned, WordPress (which I have developed) is the only one that needs a Bearer token sent as header. Google works fine for me without any patch of this kind.

Remo:
it depends on the API call you're trying to make with Google. Reading works, but POST calls don't work, more over here https://github.com/hybridauth/hybridauth/issues/389

Shade-:
Well that is not a problem now that you have the piece of code that fixes the issue.

Remo:
Yes, but I'd like to have that code in place by default so others don't need to know about it. Shall we add it to google and wordpress and leave the rest as it is till someone reports a problem?

Shade-:
Yes. The problem is, WordPress requires a **Bearer**, not an **OAuth** header which is the one required by Google in POST requests. As you can see, not only it is present in just 2 providers but it may also vary between these ones.

Also, I'm not entirely sure if the others work fine if they pass the access_token both in the querystring **and** in the header. It's safer to add to every provider once they update their APIs. Evernote, for example, specifies that tokens must be passed in the header and not in the querystring, and it returns an error if done in the latter way if I am not mistaken.

Remo:
okay, thanks for the feedback! I'm closing this, this PR https://github.com/hybridauth/hybridauth/pull/406 adds the Google header which was required for me to get things working

PR #394, opened 2015-02-16
Created WordPress and Dribbble provider

Shade-:
This is a preliminary (and the simplest) integration with WordPress and Dribbble, however they both work fine in my tests. There is still something to do (eg.: posting on blog / dribbble), but I'd say they are stable enough to be added to hybridauth.


AdwinTrave:
Thank you! If you could provide documentation into the `gh-pages` branch that would be greatly appreciated.

Shade-:
Sure thing. Will do it asap.

PR #397, opened 2015-02-18
Rrestore broken functionality: with this changes we provide for users to ...

sidz:
With this changes we provide for users to use their own storage mechanism. We can create our Hybrid_Storage class and register it in composer.json!

Example:
````
<?php
function __autoload($class)
{
    include($class . '.php');

    // Check to see whether the include declared the class
    if (!class_exists($class, false)) {
        trigger_error("Unable to load class: $class", E_USER_WARNING);
    }
}

if (!class_exists('MyClass', false)) {
    echo 'create base';
} else {
    echo 'nope. use another';
}
?>
````
But when we remove second parameter we can register own Hybrid_Storage!

### DIFF
--- a/hybridauth/Hybrid/Auth.php
+++ b/hybridauth/Hybrid/Auth.php
@@ -1,405 +1,405 @@
 <?php
 /**
-* HybridAuth
-* http://hybridauth.sourceforge.net | http://github.com/hybridauth/hybridauth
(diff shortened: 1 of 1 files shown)


AdwinTrave:
Good idea, but looks like your editor changed the file formatting so its impossible to distinguish what are the specific changes you made, so I have to deny this. Please try to resubmit again without rewriting the entire file. Thanks!

PR #399, opened 2015-02-21
MailChimp OAuth2 provider

groganz:


AdwinTrave:
Thank you! If you could provide documentation into the `gh-pages` branch that would be greatly appreciated.

PR #403, opened 2015-02-28
Updated Dribbble docs (Disqus ID missing)

Shade-:
Sorry, I forgot to update some things in the new documentation.

Anyway, I would like to ask you a question: how is Hybridauth 3.x development going? I haven't seen recent commits and that's a shame because it looks A LOT more professional and neat than the current version.


AdwinTrave:
I agree with you. HybridAuth 3.x is developed by @miled and I currently don't have the time to do much more then maintain what is already here. So it will depend on him and if somebody else steps up to.

PR #404, opened 2015-02-28
Update Endpoint.php

j05u3:
Logger can't be called before it is included, this was throwing me 'class not found'. Including is done 3 lines after that in "initialize".


j05u3:
The same for Hybrid_Exception.

PR #405, opened 2015-03-02
Add BitBucket Provider

gabrielsomoza:


gabrielsomoza:
This is not completely ready yet, still need to actually run it and make sure it works.

gabrielsomoza:
Alright, tested it and it works! Let me know if you need anything else in order to merge this.

AdwinTrave:
Thank you! The only other thing is that you can make a documentation page with all the necessities on the `gh-pages` branch.

gabrielsomoza:
Ok - when is this scheduled to be released? Just to make sure I add the docs before that.

AdwinTrave:
When I get time to resolve some other issues like #406 and #408. Some of it depends on @miled, but I think I'll get the time 2 weeks from now to do all the work for release, but no longer then 3 weeks.

Issue #296, opened 2014-08-20
OAuth2Client - curl - invalid POST


dewa1t:
In hybridauth/Hybrid/thirdparty/OAuth/OAuth2Client.php:219 POST params are setted to curl request. They are passed as an array, and according to http://php.net//manual/ru/function.curl-setopt.php they should be passed as string.

Request, that gets access token via POST is failes.

The problem can be solved the next way:

```
diff --git a/hybridauth/Hybrid/thirdparty/OAuth/OAuth2Client.php b/hybridauth/Hybrid/thirdparty/OAuth/OAuth2Client.php
index b7f1aca..43ea183 100644
--- a/hybridauth/Hybrid/thirdparty/OAuth/OAuth2Client.php
+++ b/hybridauth/Hybrid/thirdparty/OAuth/OAuth2Client.php
@@ -216,7 +216,7 @@ class OAuth2Client
 
                if( $type == "POST" ){
                        curl_setopt($ch, CURLOPT_POST, 1);
-                       if($params) curl_setopt( $ch, CURLOPT_POSTFIELDS, $params );
+                       if($params) curl_setopt( $ch, CURLOPT_POSTFIELDS, http_build_query($params) );
                }
 
                $response = curl_exec($ch);

```

PR #406, opened 2015-03-02
add access token to header in google provider

Remo:
* Replaces #390 
* This should make it possible to call google POST methods
* We still have to undo this change https://github.com/hybridauth/hybridauth/issues/296 as explained here https://github.com/hybridauth/hybridauth/issues/389#issuecomment-72837755


AdwinTrave:
Could you undo #296 here so it is all in one place? Thank you!

Remo:
I thought about that, but it's not that easy. If I undo that change, we'll break sites that depend on that code. Once we undo it, everyone who depends on it will have to build the query string before calling hybridauth.. Adding an option to disable the auto query string building is ugly too, but it wouldn't be a breaking change.. Any ideas?

AdwinTrave:
#296 was changed on January 3rd, so it is part of 2.4 version and not the current 2.3 hence only people who keep up with the development branch depend on it.

Remo:
Okay I see.. This OAuth stuff is confusing me more and more, now I have to use `Authorization`..

AdwinTrave:
@Remo So it should be:
```php
$this->api->curl_header = array("Authorization: OAuth " . $this->access_token);
```
???

PR #408, opened 2015-03-04
Make Hybrid_Endpoint methods non-static

nazar-pc:
Changed static methods to dynamic, also constructor added.
`::process` method left for backward-compatibility and calls constructor.

Should not break any existing code using this according to documentation.
Also big question whether we need all that methods and properties to be public?

### DIFF
--- a/hybridauth/Hybrid/Endpoint.php
+++ b/hybridauth/Hybrid/Endpoint.php
@@ -11,15 +11,15 @@
  * Hybrid_Endpoint class provides a simple way to handle the OpenID and OAuth endpoint.
  */
 class Hybrid_Endpoint {
-	public $request = NULL;
-	public $initDone = FALSE;
+	protected $request = NULL;
+	protected $initDone = FALSE;
 
 	/**
 	 * Process the current request
 	 *
 	 * $request - The current request parameters. Leave as NULL to default to use $_REQUEST.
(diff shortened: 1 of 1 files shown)


AdwinTrave:
I agree with you, but would appreciate @miled feedback on this before merge.

miled:
this class is definitely better now. thank you!

> Also big question whether we need all that methods and properties to be public?
no. only `::process` need to be public, everything else can be protected.

nazar-pc:
Just made everything except `::process()` and constructor protected

AdwinTrave:
Got an error:
```
Severity: Parsing Error
Message: syntax error, unexpected T_STATIC, expecting T_STRING or T_VARIABLE or '$'
Filename: Hybrid/Endpoint.php
Line Number: 69
```

specifically here:
```php
public static function process( $request = NULL )
{
	new static( $request );
}
```

nazar-pc:
Is that PHP 5.2?

nazar-pc:
Can't remember when I touched 5.2 last time, anyway, should work now as late static binding for modern versions, and as just `self` for 5.2

AdwinTrave:
Darm, looks like it. I fixed my hosting to 5.4, but the fix worked as well, at least we won't hear complains from legacy people.

PR #420, opened 2015-03-19
Steam API user data support

jacob1237:
Hi!

I've added Steam API support for user data retrieving to Steam OpenID provider.
Actually, it's just a little modified/refactored version of Steam provider from 3.0.0 branch of HybridAuth project.

I don't know why this patch exists only in 3.0.0 version, but I think it needs to be in current stable version too.

### DIFF
--- a/additional-providers/hybridauth-steam/Providers/Steam.php
+++ b/additional-providers/hybridauth-steam/Providers/Steam.php
@@ -1,55 +1,82 @@
-<?php 
+<?php
 /*!
 * HybridAuth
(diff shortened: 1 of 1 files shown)


AdwinTrave:
Have you tested it?

jacob1237:
I've tested only Steam API method, because I'm using it on my current project.
I can try to test the legacy method later, but maybe just remove it permanently?

The legacy method don't give any needed data about the user and sometimes return HTTP 503 error (according to Steam developer docs and my experience with this).

Issue #435, opened 2015-04-06
Google Authentication Error


rahmadimaru:
Hello, I've just build an apps and it's still in production somehow I need add a new library and do an update via composer update and it became an issue on google authentication .
Somehow I just tracking if it's happen in my wrong code because on me errors always return "Undefined property:stdClass::$emails" , but I realized it's not, the issue come from an update of hybridauth library maybe my version has been update into 2.4.0 (because as i read in documentation it's the closest update from this date/time), so I tried to rollback and back into hybridauth 2.3.0 and it solved.

![gauth bugs](https://cloud.githubusercontent.com/assets/8717152/7009433/657fb88a-dcc4-11e4-9c30-62f6585209b4.PNG)

PR #436, opened 2015-04-07
fix Undefined property: stdClass::$emails

tsukimiya:


AdwinTrave:
Is this fix for #435 ?

PR #437, opened 2015-04-07
Bugfix for Oauth 2.0

jaakon:
Without this fix the response will never be returned as object but as json string instead.

Also: emails is not always a defined property of $verified --> error in foreach (Sorry, couldn't make two separate pull requests)

### DIFF
--- a/hybridauth/Hybrid/thirdparty/OAuth/OAuth2Client.php
+++ b/hybridauth/Hybrid/thirdparty/OAuth/OAuth2Client.php
@@ -141,7 +141,7 @@ class OAuth2Client
 		}
 
 		if( $response && $this->decode_json ){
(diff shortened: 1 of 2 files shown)


hybridauth/Hybrid/thirdparty/OAuth/OAuth2Client.php
```diff
@@ -141,7 +141,7 @@ public function api( $url, $method = "GET", $parameters = array() )
 		}
 
 		if( $response && $this->decode_json ){
-			$this->response = json_decode( $response );
+			return $this->response = json_decode( $response );
```
AdwinTrave: `return` happens on line 147, why are you putting it here?

PR #458, opened 2015-05-02
Less restrictive phpunit dependency

Soullivaneuh:
Just a composer requirements improve to allow getting latest PHPUnit 4 versions.

Note: `4.6` is the latest version.


Soullivaneuh:
Branch rebased to avoid merge conflict.

ping @AdwinTrave

Issue #459, opened 2015-05-02
PSR-2: No tabs anymore


Soullivaneuh:
Related to #34.

Would be very great to get this library compliant with the "recent" PHP standards.

The first and more important issue to solve for me is to avoid using tab on code.

Please see this part of PSR-2: http://www.php-fig.org/psr/psr-2/#2.4.-indenting

Will make a PR about that to consider it.

Issue #34, opened 2012-07-13
RFC: Refactoring / 3.0 - PSR-0, Composer, and more...


EvanDotPro:
I'd like to propose that a new branch be "officially" started, possibly as a beginning to a 3.0 release.

* Upgrade to PSR-0 and (possibly) namespaces -- I can understand the argument against making 5.3 a requirement (Edit: To carify, the extra adapters do not conform to PSR-0 and are astonishingly annoying to set up autoloading for properly in the context of other projects due to the hard-coded require_once statemnts in HybridAuth.)
* Implement some coding standards
* Remove all the static usage in favor of real OOP practices
* Add official composer/packagist support
* Make it overall more friendly for integration with projects and frameworks
* Unit tests

Yes, I could fork and just start working on it, but I'd prefer to get upstream support from the project and users so that it can hopefully be a more collaborative effort if there's sufficient interest.

Thoughts?

localheinz:
Any news on this one? 

Seems long overdue!

AdwinTrave:
v3 dev is currently PSR-4 compliant

localheinz:
@AdwinTrave 

That would be [`3.0.0-Remake`](https://github.com/hybridauth/hybridauth/tree/3.0.0-Remake)?

AdwinTrave:
Yes

lorezyra:
Any plans to implement namespaces?

nazar-pc:
As written before in this thread - already implemented in experimental 3.0-Remake branch

lorezyra:
Cool. :+1:

Soullivaneuh:
Follow PSR-1/2 practices could be also a good start.

See #459 as example.

Soullivaneuh:
@EvanDotPro please see #461.

Feel free to comment it. :+1:

wheelsandcogs:
Is 3.0 currently in a usable state? i.e. is it working with a few missing providers / small bugs or is it still under heavy dev / fairly broken?

I'm trying to make a decision on whether to go with 3.0 or 2.4 on a new project, so a rough idea of stability would really help, cheers.

miled:
@wheelsandcogs 3.0 is still under heavy development and not suited for production. afaik, there's few braves people who already using it, but the decision is yours. 

For anyone interested to know, the exact status of project can be found in https://github.com/hybridauth/hybridauth/blob/3.0.0-Remake/TODO.md

PR #461, opened 2015-05-02
Apply some PSR-2 standards rules

Soullivaneuh:
This PR applies most of the PSR-2 (and 1) standards.

This was made with the great [PHP Coding Standards Fixer](http://cs.sensiolabs.org/) tool.

I put the list of applied fixers. You can see what was did here: http://cs.sensiolabs.org/#usage

Applied fixers:

 * indentation
 * parenthesis
 * trailing_spaces
 * braces
 * eof_ending
 * function_call_space
 * function_declaration
 * lowercase_constants
 * lowercase_keywords
 * method_argument_space
 * visibility

Symfony [code standards](http://symfony.com/doc/current/contributing/code/standards.html) could be also applied. This is not required but I recommend to use it. Tell me if you want it so I will update this PR (or make another one).

I also updated `.editorconfig` file and added `fabpot/php-cs-fixer` as dev dependencies to make contributor using it.

Two another things could be done after this:

 * Subscribe to [Scrutinizer CI](https://scrutinizer-ci.com/) to get informed of code quality variations. See an example here: https://github.com/Soullivaneuh/IsoCodesValidator (badges)
 * Update `CONTRIBUTING.md` file to talk about the new style guide.

This PR is a **proposition** related to #34 and #459. I'm waiting your opinions and questions.

You can obviously not accept it but I think it would be a great improvement to get PSR-2 compliant onto this project, before 3.0 stable.

Best regards

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,12 +1,11 @@
-# http://editorconfig.org/
-
 root = true
 
 [*]
 charset = utf-8
 end_of_line = lf
-indent_style = tab
 insert_final_newline = true
 trim_trailing_whitespace = true
 
-indent_size = 8 
+[composer.json]
+indent_style = space
+indent_size = 4
--- a/composer.json
+++ b/composer.json
@@ -21,12 +21,13 @@
     },
     "require-dev": {
         "phpunit/phpunit": "4.1.*",
-        "mockery/mockery": "0.9.2"
+        "mockery/mockery": "0.9.2",
+        "fabpot/php-cs-fixer": "~1.6"
     },
     "autoload": {
         "psr-4": {
            "Hybridauth\\": "src/"
-        } 
+        }
     },
     "extra": {
         "branch-alias": {
--- a/examples/example.php
+++ b/examples/example.php
@@ -9,7 +9,7 @@
 /**
 * Step 1: Require the Hybridauth Library
 *
-* If you are not using Composer, you may use the included Hybridauth PSR-4 compliant 
+* If you are not using Composer, you may use the included Hybridauth PSR-4 compliant
 * autoloader as an example.
 *
 * If you are already using another PSR-4 autoloader on your project, simply map the
@@ -25,7 +25,7 @@ include 'vendor/autoload.php'; // or include 'hybridauth_autoload.php';
 * If you're already familiar with the process, you can skip the explanation below.
 *
 * To get started with GitHub authentication, you need to create a new GitHub
-* application. 
+* application.
 *
 * First, navigate to https://github.com/settings/applications then click the Register
 * new application button at the top right of that page and fill in any required fields
@@ -38,8 +38,8 @@ include 'vendor/autoload.php'; // or include 'hybridauth_autoload.php';
 * After configuring your GitHub application, simple replace 'your-app-id' and 'your-app-secret'
 * with your application credentials (Client ID and Client Secret).
 *
-* Providers who uses OAuth 2.0 protocol (i.g., GitHub, Facebook, Google, etc.) may need 
-* an Authorization scope as additional parameter. Authorization scopes are strings that 
+* Providers who uses OAuth 2.0 protocol (i.g., GitHub, Facebook, Google, etc.) may need
+* an Authorization scope as additional parameter. Authorization scopes are strings that
 * enable access to particular resources, such as user data.
 *
 * https://developer.github.com/v3/oauth/
@@ -47,11 +47,11 @@ include 'vendor/autoload.php'; // or include 'hybridauth_autoload.php';
 */
 
 $config = [
-	'callback' => 'https://path/to/hybridauth/examples/example.php',
+    'callback' => 'https://path/to/hybridauth/examples/example.php',
 
-	'keys' => [ 'id' => 'your-app-id', 'secret' => 'your-app-secret' ],
+    'keys' => [ 'id' => 'your-app-id', 'secret' => 'your-app-secret' ],
 
-	'scope' => [ 'user:email' ]
+    'scope' => [ 'user:email' ]
 ];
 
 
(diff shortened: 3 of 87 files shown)


miled:
I wish to keep the current style. The reason why I didn't use PSR-2 is because it's a recommendation more than a standard, besides it does not bring any additional value.

V3 will be PSR-1 and PSR-4 compliant, tho.

nazar-pc:
Same for me to be honest. Even more, PSR is just recommendation from group of people that develop frameworks, even inside this group some people voted against, so this is not "standard", just recommendation, compromise of some set of representatives.

But I think it is a good idea to have some specified code style (do we have any?). Might be good to include code style config for popular IDEs into repository, so that anyone can use it out of the box.

Which code style to choose is not as important as to be consistent across project.

P.S. Scrutinizer is awesome, we should definitely start using it) I'm using it in my project and quite happy with it.

Soullivaneuh:
@miled 

> PSR-2 is because it's a recommendation more than a standard

I'm curious of why your are saying that. Can you please explain it?

I suggested PSR-2 because it's the must use standard/recommendation on recent projects. This is why it could be great to have it a the new fresh version of HybridAuth.

Refered from the [PSR-2 introduction](http://www.php-fig.org/psr/psr-2/#coding-style-guide) extract:

> The intent of this guide is to reduce cognitive friction when scanning code from different authors.

> Thus, the benefit of this guide is not in the rules themselves, but in the sharing of those rules.

To answer of this:

> besides it does not bring any additional value.

I don't agree (surprise! ;-) ). Beside the advantage of following a common standard, you have many advantages on the code itself, take a look to the doc.

For example, the indenting note: http://www.php-fig.org/psr/psr-2/#2.4.-indenting

Or another sample: Avoid using both upper-cased and down-cased keywords (`true`, `false`, `null`, `array`...) in order to keep code coherence.

@nazar-pc 

> Might be good to include code style config for popular IDEs into repository, so that anyone can use it out of the box.

> Which code style to choose is not as important as to be consistent across project.

PSR-2 is the most followed by popular IDE like PHPStorm. It's a new argument to use it. :+1: 

Note: About indenting, for example, your `.editorconfig` force 8 column space per tab, this produce a hardly readable code.

> Scrutinizer is awesome, we should definitely start using it) I'm using it in my project and quite happy with it.

Yes, but by default Scrutinizer wouldn't like you if you don't follow PSR-2, isn't it? (just a question) :-)


---

After all, thanks guy to participate to this discussion. :+1:

PrplHaz4:
@miled @nazar-pc  that is largely true for an individual developer, but the fact is this is now a community project getting contributions from all over.

PSR-2 is a simple, well understood guideline that is easy to apply and pretty thoroughly documented.  I don't necessarily agree with all of the guidelines but IMO, having a standard that is easy to follow is more important than wasting time on a holy war over spaces and tabs.

Soullivaneuh:
> that is largely true for an individual developer, but the fact is this is now a community project getting contributions from all over.

Exactly what I was trying to say. :+1: 

> I don't necessarily agree with all of the guidelines but IMO.

Yes. This is why I applied only some of the fixers for the moment. Let's Scrutinizer say what could be improved (or not) after.

Soullivaneuh:
Maybe should we call HybridAuth contributors/users to vote about this suggestion to get an global idea of the opinion?

miled:
> Maybe should we call HybridAuth contributors/users to vote about this suggestion to get an global idea of the opinion?

I don't think that would be necessary. People will most definitely vote for a familiar standard over personal preferences. so, let settle for PSR-2..

> P.S. Scrutinizer is awesome, we should definitely start using it) I'm using it in my project and quite happy with it.

I'm already using both scrutinizer and codeclimate on my fork, but tbh I prefer codeclimate cause scrutinizer tends to complaint, a lot.

https://scrutinizer-ci.com/g/miled/hybridauth/
https://codeclimate.com/github/miled/hybridauth/

nazar-pc:
> PSR-2 is the most followed by popular IDE like PHPStorm. It's a new argument to use it. :+1:

True, but I have own code style for my projects as well as public repository with code style configuration.

> Note: About indenting, for example, your .editorconfig force 8 column space per tab, this produce a hardly readable code.

Agree, 8 is too much 4 is more common and readable for PHP

> Yes, but by default Scrutinizer wouldn't like you if you don't follow PSR-2, isn't it? (just a question) :-)

To be honest I don't know. I mean is has extremely limited configuration for code style at the moment, so it does give me some unrelated patches suggestions. I've reported such an issue for them, we'll see how they will deal with that. But I'm currently reformatting code in IDE according to configured code style automatically, so, it is not a big issue for me.

@Soullivaneuh, I agree do not mix tabs and spaces, do not mix uppercase and lowercase keywords and so on. That is just what I say - consistency.

Anyway, when I'm joining any side project I'm accepting their codestyle, independently of what it is.

nazar-pc:
Also, Scrutinizer will punish more because of code duplication, wrong or incomplete PhpDoc sections and overcomplicated code parts, code style is, mostly, a matter of taste.

Soullivaneuh:
I heard and understood your arguments, you know mines. :+1: 

Let's waiting some contributors reactions on this PR, must be interesting.

nazar-pc:
I'd not expect too much reactions, although 209 people should read this message, as usual, most of them are not active.
There is @AdwinTrave who have also write access to this repository, and someone else, don't remember exactly whom. Would be good to hear what they think is they care about this topic.

danizord:
:+1:

AdwinTrave:
I'm all for using the PSR standards. It makes it easy to reference and many developers are already familiar with them. In the end any style guide is better then nothing.
@SocalNick

Soullivaneuh:
Please don't hesitate to ping me for branch rebase if change was made on the project.

Soullivaneuh:
Thanks. :+1: 

I will take a look on CONTRIBUTING later and make a PR about that.

PR #464, opened 2015-05-04
PSR-2 patch

nazar-pc:
I've added PhpStorm code style config with pre-configured PSR-2 which will be automatically used by anyone who will want to contribute into project.
Another commit is result of reformatting using PhpStorm and PSR-2 config, this is what you'll get when press Ctrl+Alt+R in PhpStorm and likely what others will use to make sure that newly written or modified code follows PSR-2 as it should.

### DIFF
--- a/src/Adapter/OpenID.php
+++ b/src/Adapter/OpenID.php
@@ -27,24 +27,24 @@ use Hybridauth\Thirdparty\OpenID\LightOpenID;
 abstract class OpenID extends AbstractAdapter implements AdapterInterface
 {
     /**
-    * LightOpenID instance
-    *
-    * @var object
-    */
+     * LightOpenID instance
+     *
+     * @var object
+     */
     protected $openIdClient = null;
 
     /**
-    * Openid provider identifier
-    *
-    * @var string
-    */
+     * Openid provider identifier
+     *
+     * @var string
+     */
     protected $openidIdentifier = '';
 
     /**
-    * Adapter initializer
-    *
-    * @throws InvalidOpenidIdentifierException
-    */
+     * Adapter initializer
+     *
+     * @throws InvalidOpenidIdentifierException
+     */
     protected function initialize()
     {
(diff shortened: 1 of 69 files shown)


src/Adapter/OpenID.php
```diff
@@ -83,130 +83,129 @@ public function authenticate()
...
-            'namePerson/last'        ,
-            'namePerson/friendly'    ,
-            'namePerson'             ,
-            'contact/email'          ,
-            'birthDate'              ,
-            'birthDate/birthDay'     ,
-            'birthDate/birthMonth'   ,
-            'birthDate/birthYear'    ,
-            'person/gender'          ,
-            'pref/language'          ,
```
miled: there goes my evening :d

Soullivaneuh:
@nazar-pc 

Adding a vendor IDE config on a public project is not recommended at all. BTW, if contributors want to edit PHPStorm settings, this could override this file and be committed accidentally.

You have a .editorconfig. Most a the IDE (including PHPStorm) has plugins for that or built-in functionality.

BTW for the code patch, I didn't run all the fixers on #461. To be sure, you can run it like this:

```bash
php vendor/bin/php-cs-fixer fix --level=psr2 .
```

Regards.

Soullivaneuh:
With `--verbose` options to know what php-cs-fixer does.

nazar-pc:
@Soullivaneuh, I know that adding config directory is not recommended, but there is only one file in there, and it really works great, for instance, in ownCloud project. While you're contributing to HybridAuth, you shouldn't change codestyle settings (ideally).

Soullivaneuh:
@nazar-pc As you want we will see, but it's not recommended (I insist :-P).

You should instead configure Scrutinizer. I think you can use it to inspect code according to PSR-2 and more, it could propose some patches. :+1:

miled:
@nazar-pc I agree with Sullivan. There's no need for custom IDE configs.

PR #474, opened 2015-05-20
Added WarGaming OpenID provider.

duozersk:


duozersk:
This one if for the WarGaming.net Open ID

duozersk:
More info here - https://eu.wargaming.net/support/Knowledgebase/Article/View/90/22/what-is-openid-and-how-do-we-use-it

Issue #340, opened 2014-11-06
Explorer problem


andrea82:
I have problem on login to any social with internet explorer. It seams that lost the session.
I have tried changing address (www or without) the different is that i receive the message you cannot access directly on coma back from the social. In the first case i receive the same message before make the request to the social. Other browsers works fine.

PR #476, opened 2015-05-23
Ensure session is saved before sending response

tszming:
Sometimes session was not saved before doing a fast redirect (we use nginx/php5-fpm), should fix this issue #340 also.


miled:
thank you.

PR #477, opened 2015-05-24
Move Vkontakte fields to configuration

TrogWarZ:
1. Now you can use HybridAuth config to set fields that will be taken from [VKontakte](https://vk.com/) for [users.get](https://vk.com/dev/users.get) and [friends.get](https://vk.com/dev/friends.get)
2. Refactor code to use api_base_url from \OAuth2Client

### DIFF
--- a/additional-providers/hybridauth-vkontakte/Providers/Vkontakte.php
+++ b/additional-providers/hybridauth-vkontakte/Providers/Vkontakte.php
@@ -18,9 +18,29 @@ class Hybrid_Providers_Vkontakte extends Hybrid_Provider_Model_OAuth2
 	// default permissions
 	public $scope = "email";
 
+	// default user fields map
+	public $fields = array(
+		'identifier'  => 'uid',
+		'firstName'   => 'first_name',
+		'lastName'    => 'last_name',
+		'gender'      => 'sex',
+		'displayName' => 'screen_name',
+		'city'        => 'city',
+		'country'     => 'country',
+		'bdate'       => 'bdate',
+		'photoURL'    => 'photo_max_orig',
+		'nickname'    => 'nickname',
+		'domain'      => 'domain',
+		'timezone'    => 'timezone',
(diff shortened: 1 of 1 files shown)


nazar-pc:
This doesn't seem to be backward compatible.
Maybe, better assume old fields by default, but allow to redefine it using configuration?

TrogWarZ:
@nazar-pc, can you provide example when it breaks something? I'll fix this shortly.

But i think they are backward-compatible. PR just add some new fields, but not removes old (see code below).
All fields, that defined in \Hybrid_User_Contact but not got from Vkontakte, wil be null but they will exist.

```php
public $fiends = [
    'identifier'  => 'uid',            // Old field saved
    'firstName'   => 'first_name',     // Old field saved
    'lastName'    => 'last_name',      // Old field saved
    'gender'      => 'sex',            // Old field saved
    'displayName' => 'screen_name',    // Old field saved
    'city'        => 'city',           // New one
    'country'     => 'country',        // New one
    'bdate'       => 'bdate',          // Old field saved
    'photoURL'    => 'photo_max_orig', // Old field saved
    'nickname'    => 'nickname',       // Old field saved
    'domain'      => 'domain',         // New one
    'timezone'    => 'timezone',       // Old field saved
    'photo_rec'   => 'photo_rec',      // Old field saved
    'photo_max'   => 'photo_max',      // New one
    'home_phone'  => 'home_phone',     // New one
    'home_town'   => 'home_town',      // Old field saved
];
```

TrogWarZ:
Okay, i've change field map to make compability more clearly. Check this out, please.

nazar-pc:
Ah, sorry, looks like I understood it wrong, seems to be fine indeed, did you test it (because I didn't yet)?

TrogWarZ:
Yep, i use this code in my project and tested compability version from master and from my PR. It's good, just added new fields.

PR #480, opened 2015-05-30
Improved inline documentation for IDE support

hypeJunction:
I know you guys are working on 3.0, but in case you are interested, this adds inline documentation, without which it's impossible to navigate your lib.
It also delints the code base - sorry for the git noise. Working on an integration and need a more readable codebase.


AdwinTrave:
Thanks for the improvement!

PR #486, opened 2015-06-09
3.0.0 remake

ADmad:
Some PSR2 love. Added couple of travis builds too.


miled:
Thank you!

ADmad:
That was fast, could have at least let the builds finish :smile:

PR #492, opened 2015-06-11
Avoid isset checks.

ADmad:
Improves code readability.

Issue #485, opened 2015-06-06
[3.0] Use Symfony OptionsResolver


Soullivaneuh:
Symfony [OptionsResolver](http://symfony.com/doc/current/components/options_resolver.html) components is a powerful library for options management.

This can replace Abstract class properties by "defaults options" and the config property could be integrated on it too.

Interested? Could work a POV for demonstration.

miled:
Good idea indeed, but I'd like to keep the external dependencies for hybridauth down to zero.

Soullivaneuh:
@miled 

> I'd like to keep the external dependencies for hybridauth down to zero.

Can I have any explanation about why?

The power of composer is especially the possibility to use dependencies with ease and avoid to "reinvent the wheel".

This could be apply with option-resolver in this case, and buzz or guzzle http client for oauth request. This is not painful, instead, this will make a more reliable project because concentrated of principal task: OAuth protocols.

BTW in this case, this is not a really big dependency.

Are you neither interested for a POC?

Regards.

miled:
> concentrated of principal task: OAuth protocols.

Hybridauth has always been a tiny but *monolithic* library and the implementation of OAuth protocols is just a component of the package and what it has to offer. Further more, I believe there's more than enough PHP OAuth libraries out there, and honestly I'm not interested in making Hybridauth into a yet another one of those.

I totally agree with you on the importance of Composer and that's why we are using Packagist repository as our main medium of distribution, however - as you may know, there's a lot of people who either can't or don't use Composer on their projects, and yes, for their sake I'm willing to reinvent the wheel - or to include it into the package if necessary.

> buzz or guzzle http client

V3 can (at least in theory as I haven't done any actual test yet) work with other http clients and different sessions storage like memcached or redis thought [dependency injection](https://github.com/hybridauth/hybridauth/blob/3.0.0-Remake/src/Adapter/AbstractAdapter.php#L86-L91), so no there's no issue there.

The idea is that Hybridauth should work out the box without any external dependency, and the end users are free to use other libraries to replace the default Hybridauth components if they wish so (i.e, logger, session, http client).

> Are you neither interested for a POC?

I did look at OptionsResolver, and as I said it's pretty good piece of code, but it has 1034 LOC and it weight 89,5 Ko (more than 1/3 or the entire Hybridauth 3 repo), so normally I'd pass, however if you think that it's worth it, then a proof of concept would be welcome, but no promises tho.

I hope that answer you questions.

[...]

Issue #478, opened 2015-05-27
Use composer to load third party libs


hypeJunction:
Motivation is to avoid the duplication of third party libraries. E.g. we already have Facebook SDK and OAuth libs in our app project, and it would be handy to let composer resolve those.

ADmad:
:+1: for 3.0

miled:
v3.0 only uses LightOpenID (2014) and OAuth PHP Library (2007) and they had to be included into the package because 1) they are now abandonware - I think, and I'm maintaining them along hybridauth. 2) I had to hack into their codes for minor improvements and fixes.

Anyway, I don't think we'll be using any external library other than those two.

https://github.com/hybridauth/hybridauth/blob/3.0.0-Remake/src/Thirdparty/readme.md

ADmad:
On a related note have you considered using https://github.com/thephpleague/oauth2-client ?  Considering the fact that lot of providers are Oauth2 based it would considerably reduce the code you have to maintain.

miled:
OAuth2 specifications are fairly easy to implement and doesn't need any external library. The real pain however is with OpenID and OAuth 1/1.0a, and that's what those libraries are for.

ADmad:
Okay. By the way you could move the 3rdparty openid and oauth1 clients to separate repos of your own.

PowerKiKi:
I believe this should be closed since there is nothing to load via composer according to @miled.

PR #493, opened 2015-06-12
PSR3 logger

nazar-pc:
#### Problem
Since we are trying to keep Hybridauth standalone, PSR3 logger interface was not includes into source code. This would cause problems when Composer's version will be used with Hybridauth and will make bundled autoloader much more complex.

That is why bundled logger interface is only forward-compatible with PSR3 logger interface.

What does this mean? Well, it contains some of methods from PSR3's Logger interface with exactly the same names and set/type of arguments.

While this is great, PHP doesn't allow you to inject some object that doesn't directly (or by some of its parents) implement required method, which makes it impossible to inject PSR3 loggers directly.

#### Solution
For Composer we'll use dummy logger interface that will extend `\Psr\Log\LoggerInterface` and thus will be also compatible with any PSR3 autoloader!
That is why we depend on `psr/log` in `composer.json` now and have 2 autoload elements there.

#### Summary
Implementation is a bit tricky, but we achieved what we need:
* completely standalone implementation that is forward-compatible with PSR3
* with Composer we use native PSR3 logger interface instead of built-in to allow using any PSR3-compatible logger

P.S. Sorry about long explanation, but I wanted to make clear what is going in this PR.

### DIFF
--- a/src/Adapter/AbstractAdapter.php
+++ b/src/Adapter/AbstractAdapter.php
@@ -1,4 +1,5 @@
 <?php
+
 /*!
 * Hybridauth
 * https://hybridauth.github.io | https://github.com/hybridauth/hybridauth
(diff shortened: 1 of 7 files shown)


ADmad:
I don't really like this. It adds unnecessary complexity and uglifies the code. It can be as simple as having `Hybridauth\Logger\Logger` extend `Psr\Log\AbstractLogger` and implement the `log()` method (which is basically the currently `Hybridauth\Logger\Logger::_write()` method).

nazar-pc:
Agree, but for that we need to include `psr/log` into source code which I've mentioned under **Problem** title above.
Otherwise we need to use Composer exclusively, while we want to keep library standalone and usable even without it.
So, this is the best compromise I found at the moment, I'm open to other suggestions.

ADmad:
Composer is the defacto dependency manager for PHP now and I don't see any problem makes it's use a requirement.

Plus if we you really want to provide a standalone solution, release archives can be provided which include all dependencies (we do that for CakePHP).

miled:
@nazar-pc most of your PR is on point, but I don't agree on the psr3 implementation. logger is a extra and not a requirement, and requiring an external lib just to write a simple string to file doesn't make much sense to me.

I think an alternative solution would be to create a wrapper that implements hybridauth LoggerInterface and expect the end user to feed it a Psr3 compliant Logger.

```php
class PsrLogger implements LoggerInterface{
    protected $logger;

    public function setLogger(\Psr\Log\LoggerInterface $logger){
        $this->logger = $logger;
    }

    public function info($message){
        $this->logger->info($message, []);
    }

    public function debug($message, $context = null){
        $this->logger->debug($message, (array) $context); //< needs more work to interpolate the contex
    }

    public function error($message, $context = null){
        $this->logger->error($message, (array) $context);
    }
}
```

usage:

```php
class ActualLogger implements Psr\Log\LoggerInterface{
    use Psr\Log\LoggerTrait;

    public function log($level, $message, array $context = array()){
        // write to file or something
    }
}

$psrlogger = new Hybridauth\Logger\PsrLogger();

$psrlogger->setLogger( new ActualLogger() );

$adapter = new Hybridauth\Provider\Github( $config, null,  null, $psrlogger );
```

@ADmad 

> Plus if we you really want to provide a standalone solution, release archives can be provided which include all dependencies and don't require using composer (we do that for CakePHP).

That's a sound idea.

Honestly I was pushing for a standalone solution because at this point all the necessary components are already implemented in v3.0, and end users still have the option to use external libraries for the Http client, session storage and logger if they want/have to.

I had a discussion with @Soullivaneuh https://github.com/hybridauth/hybridauth/issues/485#issuecomment-109764071 on a similar subject where I gave opinion, but I want to give more reasons to why I think it would be better for the 2 libs we spoke about in #478 to be included in upstream:

- both LightOpenID and OAuth PHP should have access to hybridauth facilities like the logger.
- LightOpenID have [its own implementation of http requests based on curl and php streams](https://github.com/hybridauth/hybridauth/blob/3.0.0-Remake/src/Thirdparty/OpenID/LightOpenID.php#L196-L291), and I'm planning to make it use hybridauth Http client at some point in time.
- v3.0 only uses a small subset of original OAuth PHP library (~100 LOC) and I removed everything that hybridauth don't need. 

ps; btw, this PR is also conflicting with #492 :)

ADmad:
@miled I like your solution of adding a `Hybridauth\Logger\PsrLogger` class even though it does add an extra wrapper class. 

In general your aim to keep hybrid auth standalone and simple while still allowing to use alternate implementations for logger, http client etc. is pretty reasonable. Just need to be careful it doesn't result in reinventing the wheel.

nazar-pc:
OR, I'll change PR accordingly a bit later

nazar-pc:
Now we have wrapper instead, what you think?

miled:
looks good, thank you.

PR #496, opened 2015-06-16
Turn off notification emails

ADmad:


nazar-pc:
Can you add description why? Doesn't seem a good thing, it was especially designed to continuously control tests execution.

ADmad:
@nazar-pc Because those emails are an annoyance for the patch author. The build success / failure can be easily seen on github ui status notifications. If the author fails to notice failures the maintainer can just ping him to fix.

ADmad:
Also often the builds fails due to no fault of author like this one https://travis-ci.org/hybridauth/hybridauth/jobs/67060386

nazar-pc:
Yes,  it can be seen, but since build failed - you, as patch author, should be aware about to fix it, maintainer may just ignore patch until you fix it.
This is just one email per pull request, I'm getting emails about all your and other's pull requests just because I'm watching this repository, I'm not even a maintainer.

That build was an error, do you have permission to re-run it? It was clearly not your fault.

ADmad:
> Yes, it can be seen, but since build failed - you, as patch author, should be aware about to fix it, maintainer may just ignore patch until you fix it.

As I pointed out the failure could be a false positive with nothing to be actually fixed by the author.

> This is just one email per pull request,

From this repo yes, but for someone who contributes to multiple repos they all add up.  Please in general people get enough unneeded mails already from various sources :)

> That build was an error, do you have permission to re-run it?

Only this repo's maintainers have permission to re-run the build.

nazar-pc:
> As I pointed out the failure could be a false positive with nothing to be actually fixed by the author.

This is not a rule, we can do nothing here, unfortunately.
> From this repo yes, but for someone who contributes to multiple repos they all add up. Please in general people get enough unneeded mails already from various sources :)

I'm getting up to hundred emails every day in few mailboxes (excluding junk mails).
I have more than 50 folders only in one mailbox and one or more filters for each folder.
As the result - I have absolute majority of emails sorted into specific directories automatically, so that I can prioritize their handling.

You can add filter that will delete unneeded emails (like from Travis CI about Hybridauth pull request) right away. But this is certainly not the thing maintainers will want to have here.

PR #507, opened 2015-07-12
rewrite syntax into an equivalent way that is compatible with older PHP

daveagp:
The previous version of Logger.php was not compatible with older versions of PHP (my server runs 5.3.3, for reasons beyond my control). It gave this error:

    PHP Parse error:  syntax error, unexpected T_OBJECT_OPERATOR, expecting ')' in     
    .../hybridauth/hybridauth/Hybrid/Logger.php on line 44

(and again on lines 63 and 81)

This fix simply changes that "(new DateTime)->format(DATE_ATOM)" to two separate lines in order to work, but it has exactly the same functionality. I have tested it and it works.

Hybridauth is awesome! Thanks for such a great tool.

### DIFF
--- a/hybridauth/Hybrid/Logger.php
+++ b/hybridauth/Hybrid/Logger.php
@@ -38,10 +38,11 @@ class Hybrid_Logger {
 	 */
 	public static function debug($message, $object = null) {
 		if (Hybrid_Auth::$config["debug_mode"] === true) {
+                        $dt = new DateTime;
 			file_put_contents(Hybrid_Auth::$config["debug_file"], implode(' -- ', array(
 				"DEBUG",
 				$_SERVER['REMOTE_ADDR'],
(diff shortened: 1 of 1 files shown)


nazar-pc:
Well, 5.3 is not supported even by PHP developers...
You definitely should upgrade to 5.6.

daveagp:
I agree but unfortunately the server is running CentOS 5 which only gets as high as PHP 5.3.3. I would certainly love to not have to deal with such arcane syntax but it's not an option. Accepting this pull request just means that I don't have to keep track and update my own branch.

nazar-pc:
Since master branch seems to support PHP >= 5.2, this PR should be accepted, do not worry.

daveagp:
That's great! Yeah I had never previously encountered any problems.

PR #518, opened 2015-08-04
Hybrid Auth for DUMMIES => consistent example.

jgtoriginal:
I've been trying to implement hybridauth on a vanilla app and it took me long.
Actually docs are not consistent and the example given in this repo is not out of the box.
So after I got my stuff running I wanted to share it to the main repo so as people coming after me can save their time and implement this out of the box.

Changes made:
1) need to check if 'hauth_start' or 'hauth_done' have been requested in order to get Hybrid_Endpoint::process() running properly.
2) if none is requested then we can proceed and get our data out of the user.
3) $config needs to be setted up as a variable otherwise we can't get hold of it on methods further down.

### DIFF
--- a/hybridauth/index.php
+++ b/hybridauth/index.php
@@ -9,7 +9,20 @@
 //	HybridAuth End Point
 // ------------------------------------------------------------------------
 
+$config = require 'config.php';
 require_once( "Hybrid/Auth.php" );
 require_once( "Hybrid/Endpoint.php" );
 
-Hybrid_Endpoint::process();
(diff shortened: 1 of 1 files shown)


AdwinTrave:
@miled What is your opinion on this?

miled:
sorry for the delay. 

index.php is the default callback page and it should be kept as it is.

jgtoriginal:
@miled can you please then provide an alternative solution to this problem then?
I appreciate you got some design concerns, but we need to find a balance between the former and the latter. Bare in mind that because of such silly problems, many people would not use this implementation, which wouldn't help to make this more popular.
Thx!!

PR #523, opened 2015-08-11
[Draugiem] API does not always return the birthday object

MatissJanis:
Fixed an error caused by API not always returning the birthday of the user.


AdwinTrave:
Thank you!

PR #525, opened 2015-08-12
[Facebook] Fixes getUserProfile v2.4 compatibility

krileon:
Facebook changed profile data requests to provide the minimal of name and id. Without specifying the fields HybridAuth uses it'll only have those two for their profile data. This fixes it to request the fields used by HybridAuth specifically.


AdwinTrave:
Thank you!

PR #549, opened 2015-09-21
Update Auth.php library version

isantolin:


AdwinTrave:
Forgot about that. Thanks!

PR #559, opened 2015-10-05
Fix warnings for getUserByResponse() in Vkontakt provider.

ShlyapkinG:
This method doesn't consider that user can have empty fields as 'country' and 'city'. So there are will be two warnings.


AdwinTrave:
Thanks!

merkuriy:
In my opinion there is something wrong `is_string($city)` and `property_exists($city, 'name')`.
Perhaps it would be better just like so #565

AdwinTrave:
@merkuriy Feel free to submit a new PR.

PR #560, opened 2015-10-05
Require OAuth.php only if not exist

kgaut:
I had problems with the following combination :

Drupal + Drupal HybridAuth + Drupal OAuth + HybridAuth because OAuthConsumer class is already provided within Drupal OAuth.

=> Callback from linkedin and twitter lead to this error message : 

```
PHP Fatal error:  Cannot redeclare class OAuthConsumer in /var/www/sites/domain/sites/all/modules/oauth/lib/OAuth.php on line 20
```

@see : https://www.drupal.org/node/2147429


AdwinTrave:
Not a HybridAuth issue since the issue is caused by incompatibility with another Drupal module, but since this could help others as well I'll allow this in.

PR #576, opened 2015-10-29
Improve logging and error handling in main providers

FractalizeR:
With this PR I propose to improve HybridAuth logging and error reporting by the following:

 - Where appropriate, propagate nested exception
 - Where appropriate, output actual server response in "invalid response"-like messages to make debugging easier.

### DIFF
--- a/hybridauth/Hybrid/Logger.php
+++ b/hybridauth/Hybrid/Logger.php
@@ -88,4 +88,15 @@ class Hybrid_Logger {
 		}
 	}
 
+    /**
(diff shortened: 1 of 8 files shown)


StorytellerCZ:
I think the more detailed reporting should be only when debug mode is turned on.

FractalizeR:
@StorytellerCZ You mean response contents, right? Because nested exception should be propagated in all cases, I believe. It eases problem discovery on production servers using tools like Kibana.

PR #582, opened 2015-11-13
Facebook login read_stream deprecated for 2.5

neo22s:
For current facebook api (v2.5) read_stream permission is deprecated and not available for new apps. This means that facebook login cannot be used for new sites. (More info here: https://developers.facebook.com/docs/apps/versions)


Shade-:
Guys, this is CRUCIAL for Facebook's login. Otherwise you get an annoying Exception "Facebook returned an invalid user id" with code 5 from HybridAuth if you use the read_stream permission, which is indeed deprecated from v2.4 of FB's APIs.

This should be merged and released urgently.

neo22s:
Yep...also pretty simple mod as you can see ;)

Shade-:
I literally run into a brick wall trying to solve this issue for one of my products, since neither Facebook or HybridAuth return a descriptive error message.

tchalvak:
Had an identical problem with this deprecated permission.

PR #598, opened 2015-12-23
Update GitLab API points, fix LastFM typo, refactor file/folder structure stack exchange

christ0ph3r:
<h3>GitLab changes</h3>

html_url does not exist in GitLab API from what I have read on their docs.  html_url is a GitHub scope so maybe got in here because of that or GitLab supported it previously.

Location does not exist in GitLab API user endpoint.

website_ur needs to be website_url ( looked like a typo )

Docs reference http://doc.gitlab.com/ce/api/users.html

<h3>LastFM changes</h3>

Fix provider wrapper name

<h3>Stack Exchange</h3>

Change file/folder structure to be like all other additional providers.  So now it has its own folder and is not sitting alone.  Also added .php extension to file name.


StorytellerCZ:
Thanks!

PR #610, opened 2016-01-22
Fixes redirect in auth after user pushes the browser back button

nullziu:
If after a successful login a user pushes the back button on his browser, the redirect in Auth.php gets called but the url passed to it is null because it was removed from the session after the user was logged in, the end result being a blank page.

### DIFF
--- a/hybridauth/Hybrid/Provider_Adapter.php
+++ b/hybridauth/Hybrid/Provider_Adapter.php
@@ -281,6 +281,12 @@ class Hybrid_Provider_Adapter {
 		// get the stored callback url
 		$callback_url = Hybrid_Auth::storage()->get("hauth_session.{$this->id}.hauth_return_to");
 
+		// if the user presses the back button in the browser and we already deleted the hauth_return_to from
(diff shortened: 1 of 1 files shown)


nazar-pc:
I do not think this should be fixed. It is your duty not to call `::redirect()` and/or provide necessary checks in your code that might cause calling `::redirect()`.

nullziu:
@nazar-pc I think that the average user expects things to work "out-of-box" without much fuss, with that said, this pull request does not introduce any overhead, and accounts for use cases when this redirect issue is not handled properly (ie: user clicking on the back button of his browser after successful authentication).

nazar-pc:
HybridAuth works out of the box, but you have to handle certain cases yourself during integration.
Anyway, lets wait what maintainers think about this change.

nullziu:
@nazar-pc I think you totally misunderstand the point in this pull request.
 I changed my code and hopefully it clarifies the scenario.

nazar-pc:
Now I understand it better. Well, as workaround it might be the case, also throwing an exception might be a solution. Anyway it should not be as it is currently, we can't redirect to `null`, this is for sure.

PR #629, opened 2016-03-05
Adding function to list Github repositories

jossday:
New function added in hybridauth/Hybrid/Providers/Github.php
This function allows to list repositories of the user who connects
himself through hybridauth. Just use it by calling getUserRepos().

### DIFF
--- a/hybridauth/Hybrid/Providers/GitHub.php
+++ b/hybridauth/Hybrid/Providers/GitHub.php
@@ -1,0 +1,131 @@
+<?php
+/*!
+* HybridAuth
+* http://hybridauth.sourceforge.net | https://github.com/hybridauth/hybridauth
+*  (c) 2009-2011 HybridAuth authors | hybridauth.sourceforge.net/licenses.html
+*/
+
+/**
+ * Hybrid_Providers_GitHub
+ */
+class Hybrid_Providers_GitHub extends Hybrid_Provider_Model_OAuth2
+{
+	// default permissions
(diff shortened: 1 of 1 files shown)


nazar-pc:
1. This is Third party provider, why did you copy it into core providers?
2. I feel strongly against addition of new method - this is just API call, we can add multitude of them for each provider, but this is not an intention of this project

hybridauth/Hybrid/Providers/GitHub.php
```diff
@@ -0,0 +1,131 @@
...
+				throw new Exception( "User email request failed! {$this->providerId} returned an error: $e", 6 );
+			}
+		}
+
+		return $this->user->profile;
+	}
+	/**
+	*
+	*/
+	function getUserContacts() {
```
StorytellerCZ: Missing comment here.

StorytellerCZ:
As @nazar-pc said.

jossday:
Ok, sorry I only wanted to help people who could look for this.

StorytellerCZ:
@jossday No harm done :) Feel free to add it to the Github provider in [additional providers](https://github.com/hybridauth/hybridauth/tree/master/additional-providers/hybridauth-github/Providers)

nazar-pc:
Even better, create own custom provider for your project by extending class of GitHub provider.

As far as I remember this library abstracts some things across all providers, but many other things are available in individual APIs and my opinion is hat those should not be added to providers itself, because we'll end up with infinitely bloated and fragile classes. For individual purposes this might be useful though.

PR #632, opened 2016-03-24
[Facebook] Directly check the returned Facebook code for getting an access token

klausi:
When using Facebook as login provider there is a problem on multilingual sites that do additional redirects with for example "/en/" or /de/" URL prefixes. See https://www.drupal.org/node/2684015 for background.

This is the proposed solution from that issue, not sure it is correct or the way to go.


nazar-pc:
I've followed to read about issue and still do not understand how it relates to Hybridauth. I have Hybridauth integrated into my system and even with prefixed multilingual paths have no issues. It seems like Drupal-specific problem, not generic one.
If Drupal does some unnecessary redirects - force it not to do that, there is nothing Hybridauth should fix in this case.

PR #637, opened 2016-04-21
Fix fetch email from new BitBucket API

jeanbez:
The URL to fetch email was changed in version 2 of the API, thus no email was returned. This fixes it.


StorytellerCZ:
Thank you!

PR #641, opened 2016-04-30
Added actual coding standards

lsolesen:
I've added the actual coding standards based on PSR2 that this library is coded after. If accepted we could fix some of the coding standard issues quite easily with `phpcbf`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,20 @@
+language: php
+
+php:
+  - 5.3
+  - 5.4
+  - 5.5
+  - 5.6
+  - 7.0
+  - hhvm
+
+install: travis_retry composer install --no-interaction --prefer-source
+
+script:
+  - vendor/bin/phpcs --standard=phpcs.xml hybridauth
+  - vendor/bin/phpcs --standard=phpcs.xml additional-providers
+
+matrix:
+  allow_failures:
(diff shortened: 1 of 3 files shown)


StorytellerCZ:
I'm all for this, but will have to defer to @miled for the final OK.

nazar-pc:
I'd say all stuff of this kind should only target 3.x branch. And 3.x should be fixed as alpha release and finally moving towards stable. Otherwise 2.x diverges further and it would be very difficult to bring everything back.

lsolesen:
I think the commit can safely be added. It does not break anything and test
will pass as the phpcs.xml reflects the current coding standards. That will
help not introduce other violations. Then in 3.x branch we can remove the
modifications to the psr2 coding standard and get 3.x closer to psr2. If
you accept and merge the pull request into 3.x branch also.I will help fix
up the violations.

Den man. 2. maj 2016 02.42 skrev Nazar Mokrynskyi <notifications@github.com

StorytellerCZ:
I plan to release 2.7 some time this month. If we don't hear from @miled soon I will merge this. @lsolesen Feel free to go ahead and create PR for the 3.0 branch with stricter rules.

StorytellerCZ:
Will include this in v2.9

PR #642, opened 2016-04-30
Fixing issues pointed out by Sensio Insight

lsolesen:


StorytellerCZ:
@lsolesen Can you please update this. Then it is good to go. Thanks!

ApacheEx:
- 1+ year of inactivity
- lots of code changes since v2.7.0
- there is already quite good v3 which has more priority for support.

PR #652, opened 2016-05-30
New QQ Provider.

holmescn:
The QQ provider is reimplemented. The original one add many custom function that helps get the user id and user profile. I have rewritten the provider based on github provider. No additional function are needed, all request sent and retrieved by OAuth2Client class.

I have no idea about whether I should add a new provider instead of override the original one.

### DIFF
--- a/additional-providers/hybridauth-qq/Providers/QQ.php
+++ b/additional-providers/hybridauth-qq/Providers/QQ.php
@@ -1,35 +1,18 @@
 <?php
 /*!
 * HybridAuth
-* http://hybridauth.sourceforge.net | http://github.com/hybridauth/hybridauth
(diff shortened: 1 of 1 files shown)


StorytellerCZ:
I'm mostly hesitant as this breaks existing functionality.

Could you re-implement the old functions and add depreciated log message? Then in 2.8 or 2.9 we can remove the old functionality.

holmescn:
The original implementation just do redundant works that HybridAuth already done. I don't think anything should be marked as depreciated because they are different implementation. If I do it, perhaps everything in the original one will be marked deprecated, and should be eliminated in the next version.

christ0ph3r:
Found 1 spelling error 

deocde

christ0ph3r:
We should always try out best not to break things for others.

holmescn:
I thought the best way is to make another PR that rename my implementation.

PR #656, opened 2016-06-21
Duplicate conditional expression in Sina.php

kilida:
One of these expressions can be removed, unless it was supposed to be `https`?


christ0ph3r:
Nice find!  Yeah that duplicate does not need to be there.

Looks good to me.  Good to merge.  @StorytellerCZ

PR #658, opened 2016-07-01
Workaround for Windows Live authentication

mourjan:
A while ago, Microsoft enforced their oauth security by disallowing redirect url to contain any parameters.
Hence the following redirect url does not work:
http://mywebsite.com/path_to_hybridauth/?hauth.done=Live

A workaound created by @tohweesiang was to provide another redirect url format in the case of Live authentication, so the redirect url will become like:
http://mywebsite.com/path_to_hybridauth/live.php

If this pull request is accepted than [Hybridauth's documentation for Windows Live](http://hybridauth.sourceforge.net/userguide/IDProvider_info_Live.html) should be changed as well,

regards

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "hybridauth/hybridauth",
+    "name": "mourjan/hybridauth",
     "description": "Open source social sign on PHP library.",
     "keywords": ["hybridauth", "login", "oauth", "social", "openid", "facebook", "google", "twitter", "yahoo"],
     "homepage": "http://hybridauth.sourceforge.net",
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "mourjan/hybridauth",
+    "name": "hybridauth/hybridauth",
     "description": "Open source social sign on PHP library.",
     "keywords": ["hybridauth", "login", "oauth", "social", "openid", "facebook", "google", "twitter", "yahoo"],
     "homepage": "http://hybridauth.sourceforge.net",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "hybridauth/hybridauth",
+    "name": "mourjan/hybridauth",
```
StorytellerCZ: Please fix this back.

mourjan: Done..

am I supposed to submit another Pull Request?

Bassel

hybridauth/Hybrid/Auth.php
```diff
@@ -352,6 +352,9 @@ public static function logoutAllProviders() {
 	 * @param string $mode PHP|JS
 	 */
 	public static function redirect($url, $mode = "PHP") {
+		if(!$mode){
+			$mode = 'PHP';
```
christ0ph3r: I am not sure if this is needed.   Can you explain why it is?

```
<?php 

function redirect($url, $mode = "PHP") {
  echo $mode;
}

redirect('https://test.com/');
?>
```

will output PHP because the second parameter is already set 

$mode = "PHP"

so I do not understand why check it again.

christ0ph3r: Nevermind.  I understand now.  So incase its empty this is fallback.

mourjan: Exactly, in some cases **$mode** is passed as an empty string, and in this case **$mode** will not be initialized by '_PHP_' value since it is not **null**

composer.json
StorytellerCZ: Just pushing to this branch should do.

PR #673, opened 2016-09-05
Add strava as new provider

rittresultater:
Strava is the social network for athletes. A global community of millions of runners, cyclists and triathletes, united by the camaraderie of sport.


StorytellerCZ:
Thanks a lot!

PR #684, opened 2016-10-05
Set emailVerified for Twitter provider

rbayliss:
If an e-mail address is returned by Twitter, it's considered verified as per their documentation of the verify_credentials endpoint.


StorytellerCZ:
Good catch! Thanks a lot!

PR #685, opened 2016-10-07
Remove array shorthand constructor from LinkedIn provider

ragulka:
Shorthand arrays are PHP 5.4+, however the rest of this lib seems to be aimed for 5.2+, so I've replace the shorthand version with the usual `array()`.


StorytellerCZ:
Given the EOL for PHP 5.2 it shouldn't be any concern, but let's be consistent.

Thanks a lot!

PR #634, opened 2016-04-04
Google plus api correct age values

driade:
As seen in real world apps, it's possible not to receive $min, so Hybridauth throws an Exception.

Didn't find much info about it, but this page explaining the Java class for Android gives clues about the issue

https://developers.google.com/android/reference/com/google/android/gms/plus/model/people/Person.AgeRange#public-methods

So, we cannot take for granted we'll receive $min.

Here you can see the typical Exception you may get

```
[2016-04-04 21:36:48] production.ERROR: exception 'ErrorException' with message 'Undefined property: stdClass::$min' in /var/www/lektu/vendor/hybridauth/hybridauth/hybridauth/Hybrid/Providers/Google.php:166
Stack trace:
#0 /var/www/lektu/vendor/hybridauth/hybridauth/hybridauth/Hybrid/Providers/Google.php(166): Illuminate\Exception\Handler->handleError(8, 'Undefined prope...', '/var/www/lektu/...', 166, Array)
#1 /var/www/lektu/vendor/hybridauth/hybridauth/hybridauth/Hybrid/Provider_Adapter.php(228): Hybrid_Providers_Google->getUserProfile()
#2 /var/www/lektu/app/routes.php(3284): Hybrid_Provider_Adapter->__call('getUserProfile', Array)
#3 /var/www/lektu/app/routes.php(3284): Hybrid_Provider_Adapter->getUserProfile()
#4 [internal function]: {closure}()
#5 /var/www/lektu/bootstrap/compiled.php(5398): call_user_func_array(Object(Closure), Array)
#6 /var/www/lektu/bootstrap/compiled.php(5065): Illuminate\Routing\Route->run(Object(Illuminate\Http\Request))
#7 /var/www/lektu/bootstrap/compiled.php(5053): Illuminate\Routing\Router->dispatchToRoute(Object(Illuminate\Http\Request))
#8 /var/www/lektu/bootstrap/compiled.php(715): Illuminate\Routing\Router->dispatch(Object(Illuminate\Http\Request))
#9 /var/www/lektu/bootstrap/compiled.php(696): Illuminate\Foundation\Application->dispatch(Object(Illuminate\Http\Request))
#10 /var/www/lektu/vendor/barryvdh/laravel-debugbar/src/Middleware/Stack.php(34): Illuminate\Foundation\Application->handle(Object(Illuminate\Http\Request), 1, true)
[...]

PR #691, opened 2016-10-20
[Google+] Correct age range

driade:
As of pull request #634 , finally correct how we manage Google+ ages.

This could be a breaking change as we modify the way the data is returned and it's possible users of Hybridauth tend to parse the output.

Thanks!


driade:
I'm very sorry my branch was not synced.

StorytellerCZ:
No problem, thanks for the work! I will mark this as a breaking change in the changelog so people know where to look just in case.

PR #697, opened 2016-10-31
Modifying of Tumblr function for posting.

qzmenko:
I changed the function setUserStatus in provider Tumblr. Now you can post not just a string with the text type, but other types of blog posts (see https://www.tumblr.com/docs/en/api/v2#posting).


StorytellerCZ:
Thanks a lot!

PR #710, opened 2016-12-09
Fixed error message delivery

mruch:
Do not pass whole exception object as a message because it contains execution stack which
is sensitive information.


StorytellerCZ:
Thanks a lot!

PR #729, opened 2017-01-27
[Facebook] Fix wrong country data

Tomicapo:
The region info does not always have the same amount of info. That means
the index $regionArr[1]  is not always the country. I changed it to
fetch the last value of the array wich will most surely be the country.


StorytellerCZ:
Thanks!

driade:
@StorytellerCZ @Tomicapo May still be wrong?

If I use Facebook Graph API to ask for 104097519627871?fields=id,location,about,name I get

```
{
  "id": "104097519627871",
  "location": {
    "city": "Alcorcón",
    "country": "Spain",
    "latitude": 40.35,
    "longitude": -3.8333333333333,
    "zip": "28921–28925"
  },
  "name": "Alcorcón"
}
```

Is this compatible with this line at Facebook.php?

`$this->user->profile->region = (array_key_exists("location", $data) && array_key_exists("name", $data['location'])) ? $data['location']["name"] : "";`

PR #732, opened 2017-01-31
* [Amazon] Fix wrong lowercase in filename for AmazonOAuth2Client cla…

aWuttig:
…ss which leads to an exception when require_once was called in `Hybrid_Providers_Amazon`


StorytellerCZ:
Thanks!

PR #733, opened 2017-01-31
* [Live] Fix authorize_url for windows live

aWuttig:
* [Live] Sets redirect_uri to redirect_uri from config array to fix problems with windows live oauth query param restriction


StorytellerCZ:
Thanks!

Issue #727, opened 2017-01-26
Facebook Graph returned an error: Error validating verification code. Please make sure your redirect_uri is identical to the one you used in the OAuth dialog request


GrafDe:
Hi.
I try to install hybridauth to my website. All providers work fine (Twitter, Yandex, Vkontakte, Odnoklassniki, Mailru), only Facebook is not work and send message: 
"Facebook Graph returned an error: Error validating verification code. Please make sure your redirect_uri is identical to the one you used in the OAuth dialog request"
sometimes i get thise message on another language, but i don't know what it is the language, maybe germany:
"Facebook Graph returned an error: URL kan niet worden geladen: Het domein van deze URL is niet toegevoegd aan de domeinen van deze app. Voeg alle domeinen en subdomeinen van je app toe aan het veld Appdomeinen in de instellingen van je app om deze URL te kunnen laden."

But i don't understand what it means.
This is my settings in facebook
![_056](https://cloud.githubusercontent.com/assets/20116695/22340026/11159250-e3fd-11e6-9fd9-e2609fb6353d.png)
this is use two path ("/oauth/hybridauth2/" and /oauth/hybridauth/) because i try different version of config settings. I also try with and without substring "?hauth_done=Facebook" .

Part of config.php of hybridauth
![_057](https://cloud.githubusercontent.com/assets/20116695/22340052/296949f0-e3fd-11e6-808f-b51faf09343a.png)

When i try login from facebook i get error:
"Unspecified error.

Original error message: Facebook Graph returned an error: Error validating verification code. Please make sure your redirect_uri is identical to the one you used in the OAuth dialog request"
![_058](https://cloud.githubusercontent.com/assets/20116695/22340091/495024b4-e3fd-11e6-88cb-62360ef10fd8.png)

This is part of code of my authorithation
![_059](https://cloud.githubusercontent.com/assets/20116695/22340186/99624356-e3fd-11e6-9426-756c251ba3c0.png)

I use last version HybridAuth 2.8.3-dev.
facebook API: 2.7
php7.0

What i doing wrong? Help me, please.

aWuttig:
Hey @GrafDe 

i changed the following line and it works for me:

https://github.com/hybridauth/hybridauth/blob/master/hybridauth/Hybrid/Providers/Facebook.php#L86

`$accessToken = $helper->getAccessToken($this->params['login_done']);`

Please discuss about it.

GrafDe:
I'm happy my friend! It is works for me.

[...]

PR #734, opened 2017-01-31
* [Facebook] Fix problem which occurs during calling the Facebook Gra…

aWuttig:
…ph API to `getAccessToken()` for a specific redirectUrl

  * Adding the `login_done` uri to the  `getAccessToken()` will fix this issue
  * https://github.com/hybridauth/hybridauth/issues/727


StorytellerCZ:
Thanks!

PR #737, opened 2017-02-04
[idQ] Add idQ Provider

nicbet:


nicbet:
Documentation in a separate PR #738

PR #739, opened 2017-02-06
3.0.0 remake

JeSuisAlrick:
An accumulation of all changes that I've done on my fork of this repo. 

The most notable change is the addition of the AuthResult class that provides more information about certain operations that occur in the auth process. The main reason for this change was because Drupal, the CMS I'm using to build a site, did not have the opportunity to perform certain tasks before HybridAuth redirected to the social network for authentication. Drupal uses Symphony to do session management and it should have been allowed to redirect to the social network instead of HybridAuth.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 /vendor
 
 /examples
+/nbproject/private/
--- a/nbproject/project.properties
+++ b/nbproject/project.properties
@@ -1,0 +1,7 @@
+include.path=${php.global.include.path}
+php.version=PHP_56
+source.encoding=UTF-8
+src.dir=.
+tags.asp=false
+tags.short=false
+web.root=.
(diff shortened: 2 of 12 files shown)


miled:
If I understood you correctly, you want Drupal to actually redirect users instead of Hybridauth? 

In that case, instead of changing the code of protocols and providers adapters, it'd suffice to use [HttpClient\Util::setExitHandler($callback)](https://github.com/hybridauth/hybridauth/blob/3.0.0-Remake/src/HttpClient/Util.php#L37-L45) or to re-write [HttpClient\redirect($url)](https://github.com/hybridauth/hybridauth/blob/3.0.0-Remake/src/HttpClient/Util.php#L26) in way so Hybridauth would yield that function entirely to Drupal through a callable or something similar.

On a side note-ish, it would've been ideal if you were to split this PR into two: one for the fixes and another to discuss any possible modification to the core.

Issue #698, opened 2016-11-01
Fatal error: Class 'Facebook\Facebook' not found


nodeffect:
I get this error below if login using Facebook:
Fatal error: Class 'Facebook\Facebook' not found in C:\wamp64\www\hybridauth\Hybrid\Providers\Facebook.php on line 56 

I also notice that the Facebook folder is missing under "hybridauth/hybridauth/Hybrid/thirdparty/" ? Not sure if that's the problem. It used to be there on older version.

Using the latest version 2.8.0. I did not encounter this problem on older version.

**UPDATE: Check below for solution.**
**UPDATE: Check below for new solution for Hybridauth v2.8.2/3.**

TheJoin95:
I get the same error with the latest version of HybridAuth (2.8):
`FastCGI sent in stderr: "PHP message: PHP Fatal error:  Class 'Facebook\Facebook' not found in /var/www/connect/lib/vendor/hybridauth/hybridauth/hybridauth/Hybrid/Providers/Facebook.php on line 56" while reading response header from upstream`

I installed HybridAuth via composer, so I also get the new facebok sdk.
Is the new facebook sdk (5.4.1) stable? Or is it included correctly? 
Because the error seems to be a namespace's error.

I fixed with a downgrade a 2.7 version of hybridauth, with the older sdk.

Alxdhere:
I had such error. I use Yii 1.x and PHP 5.5.9.

Firstly I download https://github.com/facebook/php-graph-sdk and extract it in /opt/Facebook. Declare alias in Yii in config/main.php:
`Yii::setPathOfAlias('Facebook','/opt/Facebook');`

Then, download https://github.com/paragonie/random_compat and extract it in /opt/random_compat. Add in index.php:
`require_once "/opt/random_compat/lib/random.php";`

Then, in controller write:
```
if(!function_exists('hash_equals')) {
    function hash_equals($str1, $str2) {
        if(strlen($str1) != strlen($str2)) {
            return false;
        } else {
            $res = $str1 ^ $str2;
            $ret = 0;
            for($i = strlen($res) - 1; $i >= 0; $i--) $ret |= ord($res[$i]);
            return !$ret;
        }
    }
}
```

Facebook auth work fine!

nodeffect:
Hi guys, I've actually solved the problem. This version is not as straight forward as the previous versions.
You do need to get the Facebook API from https://github.com/facebook/php-graph-sdk first.

[...]

PR #760, opened 2017-03-31
Fix fatal error during Facebook provider initialization (both for composer and non composer installation)

ApacheEx:
PR with fix for #698. Works well in both cases (composer and non-composer installation of Facebook SDK)

### DIFF
--- a/hybridauth/Hybrid/Auth.php
+++ b/hybridauth/Hybrid/Auth.php
@@ -73,6 +73,7 @@ class Hybrid_Auth {
 
 		// build some need'd paths
 		$config["path_base"] = realpath(dirname(__FILE__)) . "/";
+		$config["path_vendor"] = realpath(dirname(__FILE__) . "/../../vendor");
 		$config["path_libraries"] = $config["path_base"] . "thirdparty/";
 		$config["path_resources"] = $config["path_base"] . "resources/";
 		$config["path_providers"] = $config["path_base"] . "Providers/";
@@ -353,7 +354,7 @@ class Hybrid_Auth {
 	 */
 	public static function redirect($url, $mode = "PHP") {
 		if(!$mode){
-			$mode = 'PHP';	
+			$mode = 'PHP';
 		}
 		Hybrid_Logger::info("Enter Hybrid_Auth::redirect( $url, $mode )");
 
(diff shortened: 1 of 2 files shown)


Kirancloudnix:
SHould we upgrade the facebook graph SDK, ALso please update with clear instructions

ApacheEx:
@Kirancloudnix,

if you check composer.json - there is dependency: "facebook/graph-sdk": "^5.4". So it's already the latest version.

about PR:
1. If you clone hybridauth repository and then run "composer install":
    - you will see "vendor" directory in the root of the library which has facebook/graph-sdk and autoload.php (generated by composer)
    - so, this PR checks if there is "/vendor/autload.php" file and if so require it.
    - if somehow "vendor" directory was moved to another one, then you can provide a custom path in config.php (path_vendor) 
2. If you have hybridauth library without composer installation:
    - Facebook SDK should be placed in /hybridauth/Hybrid/thirdparty/Facebook
    - so, this PR checks if there is "/hybridauth/Hybrid/thirdparty/Facebook/autload.php" file and if so require it.

that's it. Hopefully it's quite understandable.

Kirancloudnix:
Thanks @ApacheEx

StorytellerCZ:
@ApacheEx Could you also add info about this into documentation? Thanks!

TheJoin95:
NOTE: if you run composer install in an other directory, you will have a different level for the vendor's path

dereuromark:
This actually seems to break things, see my comment @ https://github.com/hybridauth/hybridauth/issues/773 :/

Downgrading to 2.9.1 removed the fatal error for the time being.

TheJoin95:
@dereuromark Yeah, it's because you have vendor directory in some other level of directory.
You can fix this by edit the vendor variable's path with your.

dereuromark:
No, it is all default - see https://github.com/ADmad/CakePHP-HybridAuth/blob/master/composer.json
Breaking existing applications in a bugfix release is not acceptable IMO.

happyman:
⁣發送自 BlueMail ​

2017年4月16日 上午9:26, 上午9:26，在 Miki <notifications@github.com> 已写:

ApacheEx:
@dareoromark,
it seems you should change path_vendor in your hybridauth/config.php. Make sure it leads to your vendor/autoload.php, then it should work.
p.s it's not broken release, it's just wrong configuration.

sunaryohadi:
I am using custom array for config along with other component setting in my Application.

So, if I change Auth.php line 76,

`$config["path_vendor"] = realpath(dirname(__FILE__) . "/../../vendor");`
to
`$config["path_vendor"] = realpath(dirname(__FILE__) . "/../../../../../vendor");`

It run correctly.

ApacheEx:
@sunaryohadi, @dereuromark I have created a PR with fixes for autoload.php. Now, it should work for all cases. Please check it out #776

sunaryohadi:
The PR is working very well in my case .... Good job.

PR #769, opened 2017-04-08
Fix problem with Unknown Provider. because of using the incorrect var…

kaioken:
On the last update, you change a variable on getProviderConfig that cause Unknown Provider exception

Thanks


miled:
Thanks.

kaioken:
;)

PR #775, opened 2017-04-16
autoload - Windows & slash forgot?

e-motiv:
I added 2 `DIRECTORY_SEPARATOR`'s because:
1. Makes it compatible with windows server
  (I use linux server, but test everything at home with EasyPHPServer on Windows and people might use IIS in general too)
2. I'm not sure if this was a bug or typically windows again, but it seems that `__DIR__ `doesn't have a trailing directory separator, so we have to add one. If this is not the same on linux, the latter case some more coding needs to be done on the first `DIRECTORY_SEPARATOR` not to be doubled for linux


miled:
That was a bug, thanks.

PR #776, opened 2017-04-17
Improve logic for including autoload.php

ApacheEx:
Fix issue with including autload.php in v2.9.2


ApacheEx:
@StorytellerCZ,
Hi, I've also added some Installation guide based on those changes - #777, please check it out.

PR #792, opened 2017-04-23
[Stripe] Add optional stripe_landing auth parameter

IRuslan:
## Goal

For Stripe it's possible to define what type of auth window to show initially - login or register.
Currently Hybridauth library does not allow to pass such parameter.

## Description

Add support of stripe_landing option for Stripe oauth callback.
See Stripe API for details (search by 'stripe_landing' word):
https://stripe.com/docs/connect/reference


StorytellerCZ:
Can you please add info about this to documentation? Thanks!

IRuslan:
I would love to do this, but could you point me where to add? I don't see any Stripe-specific docs.

StorytellerCZ:
The docs are under a `gh-pages` branch, the specific document: https://github.com/hybridauth/hybridauth/blob/gh-pages/userguide/IDProvider_info_Stripe.html

PR #749, opened 2017-03-22
[Linkedin] OAuth 2.0 implementation

kbighorse:
using OAuth 2.0

PR #803, opened 2017-05-01
[LinkedIn] Upgrade to OAuth2

ApacheEx:
## Summary

* Breaking change
    - `setUserStatus` now accepts an associative array with keys (content, comment, visibility), for more info check [here](https://developer.linkedin.com/docs/share-on-linkedin).
    - `getUserContacts` and `getUserActivity` no longer work because LinkedIn has restricted their open API access.
    - `LinkedInException` no longer exists.

- [X] Documentation (PR #805)

## Goal

I saw PR #749 which started to migrate LinkedIn to OAuth2, but still, no progress and I perhaps many errors. So, I'd like to help and speed up migration.

## Description

This PR includes:
- Migrate provider from OAuth1 to OAuth2
- remove usages of third-party LinkedIn SDK.


klausi:
Ouch, this is a pretty heavy API break and should not be done in a minor release. Our LinkedIn Integration relies on the old XML API integration, which is now just gone. We will have to patch in the old LinkedIn files somehow when we try to upgrade.

PR #815, opened 2017-05-06
Ensure the access_token exist in the dataStore

collizo4sky:
Added an if condition to ensure the access token does exist in the data store / session before adding it to request parameters.

Not having this condition actually broke my implementation.

Harmless code change. Please merge.

cc @StorytellerCZ @miled


miled:
Thank you.

PR #836, opened 2017-05-27
apiRequest() sometimes return array hence this PHPDoc change

collizo4sky:
After parsing by Data\Parser, result could be an array. Not always an object


miled:
Thanks.

PR #841, opened 2017-06-15
add support for variable namespace alias for SREG attributes - fixes …

CybotTM:
…SREG attributes not extracted if SREG namespace attribute name is not "openid_ns_sreg"

## Summary

* Patch for #

- [x] Tested
  - [-] Unit tests
  - [-] Style
- [-] Documentation (PR # )

## Goal

Support SREG attributes if namespace attribute name is NOT "openid_ns_sreg" by replacing hard coded SREG namespace attribute name with SREG namespace detection (see getAxAttributes()).

## Description

Makes working:

openid_ns_ext1=http://openid.net/extensions/sreg/1.1
openid_ext1_nickname=sebastian.mendel

### DIFF
--- a/src/Thirdparty/OpenID/LightOpenID.php
+++ b/src/Thirdparty/OpenID/LightOpenID.php
@@ -1043,17 +1043,19 @@ class LightOpenID
     {
         $attributes = array();
         $sreg_to_ax = array_flip(self::$ax_to_sreg);
-        foreach (explode(',', $this->data['openid_signed']) as $key) {
-            $keyMatch = 'sreg.';
-            if (strncmp($key, $keyMatch, strlen($keyMatch)) !== 0) {
-                continue;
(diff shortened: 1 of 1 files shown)


ApacheEx:
from my side, it looks good, but it's a third-party library which is accessible [here](https://github.com/iignatov/LightOpenID). 
So, probably those changes should be added there and used as composer package in HA.
@miled, what do you think?

CybotTM:
The problem with the upstream is it seems abandoned. :-(

StorytellerCZ:
I agree, this should be done upstream.

So what we need to determine now:
If it is abandoned, is there a fork that is active?
Did anyone tried to contact the maintainer through other means than GitHub?
To be continued...

ApacheEx:
It's quite a small fix, so I prefer to merge it to v3.
If 3rd-party library somehow will alive again,
we will take care about it

ApacheEx:
I'm going to merge this based on https://hybridauth.github.io/license.html

PR #851, opened 2017-07-30
Add Following and Follower count for Twitter and Instagram

jojoT84:
## Summary
* Addition

- [ x] Tested
  - [-] Unit tests

## Goal
Add Following and Follower count for Twitter and Instagram

## Description
Add ability to read Following and Follower count for Twitter and Instagram.
Added two new properties for Following and Follower count  in user profile class .


ApacheEx:
The `Profile` is quite common class. It seems we need to add `data` into this class.

then you can do something like:
```php
$userProfile->data['follower_count'] = ...;
$userProfile->data['following_count'] = ...;
```

I will take care of this.

ApacheEx:
Closed in favor of #897

PR #861, opened 2017-09-06
vkontakte doesn't return email field on getProfile call

odyswise:
without setting $scape='email' function getProfile doesn't return field email from vkontakte oauth2 service

## Summary

(select one:)
* Addition

- [x] Tested
  - [ ] Unit tests
  - [ ] Style
- [ ] Documentation (PR # )

## Goal

bug fix


ApacheEx:
looks good, thanks!

PR #869, opened 2017-10-22
Add provider for OpenStreetMap, using OAuth 1.0a

jocelynj:
## Summary

* New provider

- [ ] Tested
  - [ ] Unit tests
  - [ ] Style
- [x] Documentation (PR #873  )

## Goal

Addition of OpenStreetMap authentication against v2 branch

## Description

I've added OpenStreetMap authentication on wordpress plugin, and thought it would be better to add it on hybridauth git repository instead of wordpress-social-login. I've only tested the plugin through Wordpress.

### DIFF
--- a/additional-providers/hybridauth-openstreetmap/Providers/OpenStreetMap.php
+++ b/additional-providers/hybridauth-openstreetmap/Providers/OpenStreetMap.php
@@ -1,0 +1,85 @@
+<?php
+/*!
+* HybridAuth
+* http://hybridauth.sourceforge.net | http://github.com/hybridauth/hybridauth
+* (c) 2017, HybridAuth authors | http://hybridauth.sourceforge.net/licenses.html 
+*/
+
+/**
+* Hybrid_Providers_OpenStreetMap (openstreetmap.org)
+*/
+class Hybrid_Providers_OpenStreetMap extends Hybrid_Provider_Model_OAuth1
+{
+   	/**
+	* IDp wrappers initializer 
+	*/
+	function initialize()
+	{
+		parent::initialize();
+
+		// provider api end-points
+		$this->api->api_base_url      = "https://api.openstreetmap.org/api/0.6/";
+		$this->api->authorize_url     = "https://www.openstreetmap.org/oauth/authorize";
+		$this->api->request_token_url = "https://www.openstreetmap.org/oauth/request_token";
+		$this->api->access_token_url  = "https://www.openstreetmap.org/oauth/access_token";
+
+		// turn off json parsing!
+		$this->api->decode_json = false;
+	}
+
+
+   /**
+	* load the user profile from the api client
+	*/
+	function getUserProfile()
+	{ 
+		try{  
+			$apiUrl = 'https://api.openstreetmap.org/api/0.6/user/details';
+			$response = $this->api->get( $apiUrl, array(), "text/xml" );
+		}
+		catch( Exception $e ){
+			throw new Exception( "User profile request failed! {$this->providerId} returned an error while requesting the user profile.", 6 );
+		}
+
+		// check the last HTTP status code returned
+		if ( $this->api->http_code != 200 )
+		{
+			throw new Exception( "User profile request failed! {$this->providerId} returned an error: " . $this->errorMessageByStatus( $this->api->http_code ), 6 );
+		}
+
+		$response = @ new SimpleXMLElement( $response );
+
+		$this->user->profile->identifier    = (string) $response->user["id"];
+		$this->user->profile->displayName   = (string) $response->user["display_name"];
+		$this->user->profile->description   = (string) $response->user->description;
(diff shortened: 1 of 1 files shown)


StorytellerCZ:
Thank you! Can I ask you to also provide basic documentation?

jocelynj:
Hum, I don't see where I should add basic documentation on this provider.

Can you provide an example of a provider that has this kind of documentation ?

additional-providers/hybridauth-openstreetmap/Providers/OpenStreetMap.php
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+
+   /**
+	* load the user profile from the api client
+	*/
+	function getUserProfile()
+	{ 
+		try{  
+			$apiUrl = 'https://api.openstreetmap.org/api/0.6/user/details';
```
ApacheEx: can be just `'user/details'` because `$this->api->api_base_url` will be added automatically.

additional-providers/hybridauth-openstreetmap/Providers/OpenStreetMap.php
```diff
@@ -0,0 +1,85 @@
...
+/*!
+* HybridAuth
+* http://hybridauth.sourceforge.net | http://github.com/hybridauth/hybridauth
+* (c) 2017, HybridAuth authors | http://hybridauth.sourceforge.net/licenses.html 
+*/
+
+/**
+* Hybrid_Providers_OpenStreetMap (openstreetmap.org)
+*/
+class Hybrid_Providers_OpenStreetMap extends Hybrid_Provider_Model_OAuth1
```
ApacheEx: there are no coding standards at all. Can you fix it according to PSR2 standard?

additional-providers/hybridauth-openstreetmap/Providers/OpenStreetMap.php
```diff
@@ -0,0 +1,85 @@
...
+
+		$this->user->profile->identifier    = (string) $response->user["id"];
+		$this->user->profile->displayName   = (string) $response->user["display_name"];
+		$this->user->profile->description   = (string) $response->user->description;
+		$this->user->profile->photoURL      = (string) $response->user->img["href"];
+
+		return $this->user->profile;
+ 	}
+
+	function httpRequest( $url )
```
ApacheEx: this is not used at all. Please remove.

additional-providers/hybridauth-openstreetmap/Providers/OpenStreetMap.php
```diff
@@ -0,0 +1,85 @@
...
+
+		// check the last HTTP status code returned
+		if ( $this->api->http_code != 200 )
+		{
+			throw new Exception( "User profile request failed! {$this->providerId} returned an error: " . $this->errorMessageByStatus( $this->api->http_code ), 6 );
+		}
+
+		$response = @ new SimpleXMLElement( $response );
+
+		$this->user->profile->identifier    = (string) $response->user["id"];
```
ApacheEx: it's impossible to have object and array in one instance like here `$response->user`, so pls use only array style or only object-properties style.

ApacheEx:
@jocelynj 
there is no guideline how to add documentation, but you try to find how it's done with other providers in `gh-pages` [branch](https://github.com/hybridauth/hybridauth/tree/gh-pages)

additional-providers/hybridauth-openstreetmap/Providers/OpenStreetMap.php
jocelynj: Thanks, I've fixed indentation, and functions calls

additional-providers/hybridauth-openstreetmap/Providers/OpenStreetMap.php
jocelynj: That's true - I've add several issue when making this provider, so I tried several things. The real fix was in fact to switch to a XML decoder instead of using json (which seems to be the default decoding)

additional-providers/hybridauth-openstreetmap/Providers/OpenStreetMap.php
jocelynj: I don't quite understand what you mean ?

Should I replace `$response->user["id"]` by `$response->user->id` ?

additional-providers/hybridauth-openstreetmap/Providers/OpenStreetMap.php
jocelynj: I've removed this unused function httpRequest().

jocelynj:
Do you have any example of unit tests for Providers ?

StorytellerCZ:
This is v2 so we can skip that part :(

StorytellerCZ:
@ApacheEx Looking good now?

PR #891, opened 2017-12-08
[Twitter] Remove deprecation endpoint

### LINKED ISSUE #889: Twitter - statuses/update_with_media (deprecated). We need to use update.json only
## Bug, feature or question?

Is this a bug, a feature request or a question?

Give us a short description.

### Version and provider

### Reproduction

ApacheEx:
Resolves #889


ApacheEx:
hm, Travis CI fails - but it's not related.

PR #900, opened 2017-12-26
Update Odnoklassniki.php

mihdan:
Add access_token key to params

## Summary

* Breaking change


ApacheEx:
Makes sense, thank you. Will be available in `beta5`

PR #907, opened 2018-01-06
Clean up. Add new example. Coding Standards fixes

ApacheEx:


StorytellerCZ:
I think we might want to consider moving examples into their own repository.

ApacheEx:
@StorytellerCZ, yeah it would be nice when user can check the code and try how it works at the same time.

ApacheEx:
btw, any chance to make me as collaborator here as well ?
https://github.com/hybridauth/hybridauth.github.io

we can think later how to organize examples for hybridauth.

StorytellerCZ:
I'm afraid not, not even I have access to that. We need @miled to finish the transition to full FOSS in that regard.

ApacheEx:
Do you know how to contact with him? 😅

StorytellerCZ:
@ApacheEx It's been a while since I saw @miled around.

PrplHaz4:
@ApacheEx @StorytellerCZ, @miled used to respond to the support address...might be worth a shot? hybridauth@gmail.com

PR #937, opened 2018-04-03
make google image size configurable like facebook

Ridder90:
## Summary

(select one:)

* Addition

- [ ] Tested
  - [ ] Unit tests
  - [ ] Style
- [ ] Documentation (PR # )

## Goal

My client only want profile pictures larger then 480px. This pull request makes the image size of the google provider configurable.

## Description

I took the same method as the Facebook provider. Thus adding a photo_size to the config array enables you to change the image size.

### DIFF
--- a/src/Provider/Google.php
+++ b/src/Provider/Google.php
@@ -122,7 +122,8 @@ class Google extends OAuth2
         $userProfile->emailVerified = $data->get('verified') ? $userProfile->email : '';
 
         if ($data->filter('image')->exists('url')) {
-            $userProfile->photoURL = substr($data->filter('image')->get('url'), 0, -2) . 150;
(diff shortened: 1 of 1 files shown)


src/Provider/Google.php
```diff
@@ -122,7 +122,8 @@ public function getUserProfile()
         $userProfile->emailVerified = $data->get('verified') ? $userProfile->email : '';
 
         if ($data->filter('image')->exists('url')) {
-            $userProfile->photoURL = substr($data->filter('image')->get('url'), 0, -2) . 150;
+             $photoSize = $this->config->get('photo_size') ?: '150';
```
ApacheEx: can you fix indent? (use 4 spaces instead of 5)

Ridder90: Sure! Sorry about that.

Issue #771, opened 2017-04-11
Vkontakte provider won't sign in


mkhazov:
Using latest library version 2.9.2-dev (3c1726e) I get the following error:

Vkontakte:
```
{
  "error": {
    "error_code": 5,
    "error_msg": "User authorization failed: access_token has expired.",
    "request_params": [
      {
        "key": "oauth",
        "value": "1"
      },
      {
        "key": "method",
        "value": "getProfiles"
      },
      {
        "key": "uid",
        "value": "17355012"
      },
      {
        "key": "fields",
        "value": "uid,first_name,last_name,screen_name,sex,photo_big,home_town,domain,nickname,bdate,timezone,photo_rec,domain,photo_max,home_phone,city,country"
      }
    ]
  }
}
```

Other providers such as Facebook, Twitter etc. work as expected.

ApacheEx:
The fix is available in v3.0.0-rc4.
If you want to have it in v2 - feel free to create PR.

PR #941, opened 2018-04-05
#771: Updated Vkontakte provider.

davpsh:
## Summary

* Patch for #771 

## Description

- Updated provider according [documentation](https://vk.com/dev/manuals).

- Tested manually on pure hybridauth:
screenshot

- Documentation will be updated later if PR will be merged.

### DIFF
--- a/src/Provider/Vkontakte.php
+++ b/src/Provider/Vkontakte.php
@@ -1,51 +1,78 @@
 <?php
-/*!
-* Hybridauth
-* https://hybridauth.github.io | https://github.com/hybridauth/hybridauth
-*  (c) 2017 Hybridauth authors | https://hybridauth.github.io/license.html
-*/
 
 namespace Hybridauth\Provider;
 
 use Hybridauth\Adapter\OAuth2;
 use Hybridauth\Exception\UnexpectedApiResponseException;
-use Hybridauth\Data;
-use Hybridauth\User;
+use Hybridauth\Data\Collection;
+use Hybridauth\User\Profile;
 
 /**
- * Vkontakte OAuth2 provider adapter.
+ * Vkontakte provider adapter.
+ *
+ * Example:
+ *
+ *   $config = [
+ *       'callback'  => Hybridauth\HttpClient\Util::getCurrentUrl(),
+ *       'keys'      => ['id' => '', 'secret' => ''],
+ *   ];
+ *
+ *   $adapter = new Hybridauth\Provider\Vkontakte($config);
+ *
+ *   try {
+ *       if (!$adapter->isConnected()) {
+ *           $adapter->authenticate();
+ *       }
+ *
+ *       $userProfile = $adapter->getUserProfile();
+ *   }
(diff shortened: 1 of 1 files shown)


src/Provider/Vkontakte.php
```diff
@@ -1,51 +1,78 @@
...
+     */
     protected $accessTokenUrl = 'https://api.vk.com/oauth/token';
 
     /**
-    * {@inheritdoc}
-    */
-    protected $scope = 'email';
+     * {@inheritdoc}
+     */
+    protected $scope = 'email,offline';
```
davpsh: As new change, I've added new scope, for getting access_token without expiration, because VK API doesn't provide mechanism for refreshing only access token.

They are suggest to [re-authenticate](https://vk.com/dev/authcode_flow_user?f=4.%20Receiving%20access_token):

> Get a new access_token every time it expires or if a user changes their login or password or deletes an application from the list. 

We can't refresh token, we need to get new one and this is bad solution, because we need to redirect user to external page and getting there code for exchanging on token, during processing another request:

Token validation is called inside OAuth2::apiRequest
```php
        // refresh tokens if needed
        if ($this->hasAccessTokenExpired() === true) {
            $this->refreshAccessToken();
        }
```
What that mean? For example, my token is expired and i want to get user profile, I'm loading adapter and calling getUserProfile method, where token will be validated and with suggested flow from API, we need to re-authenticate user (redirect to page to get new code/token), that's mean that getUserProfile wouldn't be executed

src/Provider/Vkontakte.php
```diff
@@ -1,51 +1,78 @@
...
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function hasAccessTokenExpired()
+	{
+		// As we using offline scope, $expired will be false.
+		$expired = $this->getStoredData('expires_in')
```
davpsh: We can't to validate by 'expires_at', because it's calculated value and it always will be less than current time at expires_in equals 0 (if pass 'offline' in scope, expires_in will be 0).

src/Provider/Vkontakte.php
```diff
@@ -56,42 +83,42 @@ protected function validateAccessTokenExchange($response)
     public function getUserProfile()
     {
         $parameters = [
-            'uid'    => $this->getStoredData('user_id'),
-            'fields' => 'first_name,last_name,nickname,screen_name,sex,' .
-                            'bdate,timezone,photo_rec,photo_big,photo_max_orig',
-            'v' => '3.0'
+            'user_ids' => $this->getStoredData('user_id'),
```
davpsh: https://vk.com/dev/users.get

src/Provider/Vkontakte.php
```diff
@@ -1,51 +1,78 @@
 <?php
-/*!
```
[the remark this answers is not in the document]
davpsh: @ApacheEx done

ApacheEx:
Many thanks. Will be released in v3.0.0-rc.4

PR #944, opened 2018-04-16
Make possible OpenID auth via POST

zorn-v:
Wargaming OpenID do post request for example

### DIFF
--- a/src/Adapter/OpenID.php
+++ b/src/Adapter/OpenID.php
@@ -82,9 +82,7 @@ abstract class OpenID extends AbstractAdapter implements AdapterInterface
             return true;
         }
 
-        $openid_mode = filter_input($_SERVER['REQUEST_METHOD'] === 'POST' ? INPUT_POST : INPUT_GET, 'openid_mode');
-
-        if (empty($openid_mode)) {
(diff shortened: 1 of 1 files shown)


src/Adapter/OpenID.php
```diff
@@ -82,7 +82,7 @@ public function authenticate()
             return true;
         }
 
-        $openid_mode = filter_input(INPUT_GET, 'openid_mode');
+        $openid_mode = filter_input($_SERVER['REQUEST_METHOD'] === 'POST' ? INPUT_POST : INPUT_GET, 'openid_mode');
```
ApacheEx: what if use `$_REQUEST` superglobal array which contains GET + POST?

```php
if (!empty($_REQUEST['openid_mode'])) {
    $this->authenticateBegin();
}
else {
    ...
}
```

zorn-v: I think you meant `empty($_REQUEST['openid_mode'])`
At all without difference, but I just left `filter_input` as it was

ApacheEx: yeah, 💯  (updated)

can you update this PR according to my comment?

ApacheEx:
Many thanks 👍

PR #948, opened 2018-04-19
Added Blizzard Battle.net Authentication

cookieman768:
## Summary

(select one:)
* New provider

- [ ] Tested
  - [ ] Unit tests
  - [ ] Style
- [ ] Documentation (PR # )

## Goal

I want to add Blizzard Battle.net authentication.

## Description

I haven't tested this thoroughly but from the testing I have done I have not found any errors.


src/Provider/Blizzard.php
```diff
@@ -0,0 +1,65 @@
...
+{
+    /**
+    * {@inheritdoc}
+    */
+    public $scope = '';
+
+    /**
+    * {@inheritdoc}
+    */
+    protected $apiBaseUrl = 'https://us.api.battle.net/';
```
Edarlingen: @cookieman768 Correct me if i'm wrong, but didn't you hardcoded region into your Provider? What about other regions?
Here is the list of available regions:
* EU
* US
* TW
* KR
* SEA

cookieman768: Ahh. I didn’t think about that. Would it be best to have a provider for each region or an option in the config? I don’t think that the other providers have this issue. Let me know what approach I should take. It would be pretty easy to change this for each region.

cookieman768: I just looked at the providers, it looks like for WeChat there is a different provider specifically for China. I’ll take that approach and make multiple providers for each region.

ApacheEx: @cookieman768, yeah `WeChat` approach looks good for me, you can do the same. So this one will be default (with US). I think we need to add EU as well. Others only if there will be request.

Edarlingen: @ApacheEx , there is only 3 regions, so I think it shouldn't be very difficult to add one extra to completely cover all available regions. You can count this as request for `APAC` region.

ApacheEx: @Edarlingen - fine for me 😏

cookieman768: I'll work on this when I get the chance, for the time being if someone really needs the other regions they can modify this version. I'll get some time in the coming weeks when I finish my finals.

src/Provider/BlizzardAPAC.php
```diff
@@ -0,0 +1,65 @@
...
+
+use Hybridauth\Adapter\OAuth2;
+use Hybridauth\Exception\UnexpectedApiResponseException;
+use Hybridauth\Data;
+use Hybridauth\User;
+
+/**
+ * Blizzard US/SEA Battle.net OAuth2 provider adapter.
+ */
+class BlizzardAPAC extends OAuth2
```
ApacheEx: just extend `Blizzard` and change this three properties:
```
/**
 * {@inheritdoc}
 */
protected $apiBaseUrl = 'https://apac.api.battle.net/';

/**
 * {@inheritdoc}
 */
protected $authorizeUrl = 'https://apac.battle.net/oauth/authorize';

/**
 * {@inheritdoc}
 */
protected $accessTokenUrl = 'https://apac.battle.net/oauth/token';
```

cookieman768: Ahh. That would be a better approach. I'll do that later today if I get the chance.

cookieman768: Oops! My bad! I completely forgot about this, I'll do it when I get some time with a computer. I've been a bit busy recently.

ApacheEx:
Thanks a lot 💯

PR #965, opened 2018-05-10
return profileURL in Instagram adapter

thedotedge:
Instagram URL did not expose profileUrl, this PR addresses that.


ApacheEx:
Thanks a lot. Make sense.

Issue #959, opened 2018-05-06
Make provider name not case sensitive


filips123:
Make option to have provider name not case sensitive. This would be good if the provider's name is located in the URL, which can cause confusion because most of the URL is in lowercase.
Note: This could be optionally.

ApacheEx:
makes sense for me. PR is appreciated.

PR #967, opened 2018-05-11
Make provider name not case sensitive

balamyt92:
| Fixed Issues?            | #959
| Major: Breaking Change?  | no
| Minor: New Feature?      | yes
| Documentation PR         |

### DIFF
--- a/src/Hybridauth.php
+++ b/src/Hybridauth.php
@@ -120,6 +120,24 @@ class Hybridauth
 
         $adapter = isset($config['adapter']) ? $config['adapter'] : sprintf('Hybridauth\\Provider\\%s', $name);
 
+        if (! class_exists($adapter)) {
+            $adapter = null;
+            $fs = new \FilesystemIterator(__DIR__ . '/Provider/');
+            /** @var \SplFileInfo $file */
+            foreach ($fs as $file) {
+                if(! $file->isDir()) {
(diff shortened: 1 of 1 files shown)


src/Hybridauth.php
```diff
@@ -120,6 +120,24 @@ public function getAdapter($name)
 
         $adapter = isset($config['adapter']) ? $config['adapter'] : sprintf('Hybridauth\\Provider\\%s', $name);
 
+        if (! class_exists($adapter)) {
```
ApacheEx: no need extra space after `!`, just `if (!class_exists($adapter))`

p.s. it was old style.

src/Hybridauth.php
```diff
@@ -120,6 +120,24 @@ public function getAdapter($name)
...
+            $adapter = null;
+            $fs = new \FilesystemIterator(__DIR__ . '/Provider/');
+            /** @var \SplFileInfo $file */
+            foreach ($fs as $file) {
+                if(! $file->isDir()) {
```
ApacheEx: the same. try: `if (!$file->isDir())`

src/Hybridauth.php
balamyt92: well, maybe fix the whole file? I was guided by the style that was already installed.

PR #986, opened 2018-07-16
Yahoo provider fix for firstname/lastname/e-mail

antiqque:
| Fixed Issues?            | `Fixed Yahoo provider` 
| Documentation PR         | 

Fixed Yahoo provider not returning e-mails, added proper scope for returning firstname/lastname/e-mail. 'sdps-r' doesn't return much useful information except uid, nickname and avatar.

### DIFF
--- a/src/Provider/Yahoo.php
+++ b/src/Provider/Yahoo.php
@@ -122,9 +122,11 @@ class Yahoo extends OAuth2
 
         // E-mail is returned only with sdpp-w scope ( Read/Write (Public and Private) )
         foreach ($data->filter('emails')->toArray() as $item) {
-            if ($item->primary) {
-                $userProfile->email         = $item->handle;
-                $userProfile->emailVerified = $item->handle;
+            $item = new Data\Collection($item);
+
+            if ($item->get('primary')) {
+                $userProfile->email = $item->get('handle');
+                $userProfile->emailVerified = $item->get('handle');
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/Provider/Yahoo.php
```diff
@@ -120,11 +120,11 @@ public function getUserProfile()
             $userProfile->gender = 'male';
         }
 
-        // I ain't getting no emails on my tests. go figures..
+        // E-mail is returned only with sdpp-w scope ( Read/Write (Public and Private) )
         foreach ($data->filter('emails')->toArray() as $item) {
-            if ($data->get('primary')) {
-                $userProfile->email         = $data->get('handle');
-                $userProfile->emailVerified = $data->get('handle');
+            if ($item->primary) {
```
ApacheEx: let's use the following:

```php
$item = new Data\Collection($item);

if ($item->get('primary')) {
    $userProfile->email = $item->get('handle');
    $userProfile->emailVerified = $item->get('handle');
}
```

antiqque: That is correct, i made the same change in my project but i forgot to add it here.

ApacheEx: @antiqque hey, any updates?

antiqque: I'm sorry but this is my first time doing this and i'm not sure what else i'm supposed to do.

PR #992, opened 2018-08-15
Map response to Data\Collection object to use the get method

Ridder90:
| Fixed Issues?            | 
| Patch: Bug Fix?          | Yes. ` Uncaught Error: Call to undefined method stdClass::get() in /vendor/hybridauth/hybridauth/src/Provider/MicrosoftGraph.php:87`
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 

Bug fix for the MicrosoftGraph provider.


ApacheEx:
this makes sense, thanks a lot 🥇

PR #995, opened 2018-08-29
Return back support for PHP 5.2

misterspelik:
| Fixed Issues?            | Yes, returned back support for PHP 5.2
| Documentation PR         | Replaced [] to array()

Changes in 4 files in Providers/ folder to make the code support PHP 5.2 again. Replaced [] construction to array() declaration.

### DIFF
--- a/hybridauth/Hybrid/Providers/Facebook.php
+++ b/hybridauth/Hybrid/Providers/Facebook.php
@@ -23,7 +23,7 @@ class Hybrid_Providers_Facebook extends Hybrid_Provider_Model {
      * @link https://developers.facebook.com/docs/facebook-login/permissions
      * @var array $scope
      */
-    public $scope = ['email', 'public_profile'];
+    public $scope = array('email', 'public_profile');
 
     /**
      * Provider API client
@@ -212,7 +212,7 @@ class Hybrid_Providers_Facebook extends Hybrid_Provider_Model {
      */
     function getUserProfile() {
         try {
-            $fields = [
+            $fields = array(
                 'id',
                 'name',
                 'first_name',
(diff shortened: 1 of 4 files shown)


misterspelik:
@ApacheEx You have a lot of places where described about PHP 5.2, at least composer.json.
All another places in the code are using array() construction.

I know about that these versions are dead but please write below and try to understand the use case please.

This branch contains the old code itself and so-so mantained. But what`s the real reason of declining this PR.
Especially when there are a lot of servers working with the old code (as our particular application) without possibility to migrate to new version of PHP due to old framework (which means to rewrite all application).

Another question is that v2 branch will not work at new PHP anyway and this is totally not needed because of 3.0 version development.

What would you suggest to do in this case?

P. S. 
Yes for sure I can use the forked repo but the goal is to use original one because of some critical fixes which might appear from community or finding something by myself and deliver to another users.

StorytellerCZ:
@ApacheEx as much as I agree with the sentiment, @misterspelik isn't wrong either.

@ApacheEx I think we should probably merge this and release a patch version for the 2.x branch since 3.x is still in RC and release 3.x branch ASAP so that we can close support for 2.x for good and put even more pressure on people to migrate. Let's set a date for this.

@misterspelik seriously though update your app. If this is a production app, it should not run on that old PHP.

ApacheEx:
Ok, guys. Thanks for explanation. 

p.s. To release v3 we need to fix/improve documentation which is the hardest part 😥 Would be nice if somebody can help.

misterspelik:
@ApacheEx I think I can participate to this process somehow. I will contact you directly about that

PR #1008, opened 2018-10-21
Update Blizzard provider to work with their new API

ErikJson:
| Major: Breaking Change?  | Yes
| Documentation PR         | No

Blizzard have opened up their new developer portal, and the old one will be discontinued on January 6, 2019. More information can be found [here](https://us.battle.net/forums/en/bnet/topic/20769317376), but the gist of it is that the base URL is changing, as well as the route for fetching the authenticated user's information has changed. Also, app developers will need to set up new app credentials on a regular Battle.net account.

I have already implemented these changes at work, and they are getting the job done for our Battle.net login. However, for this upstream change I wasn't sure about what procedure would work best; creating additional providers, or updating the current one.

In this PR I did opt for updating the current one, as this is most likely the desired solution after the old API is obsolete. It will however not work with the deprecated API, and is thus a breaking change. Please advise what you think is the best approach.

### DIFF
--- a/src/Provider/Blizzard.php
+++ b/src/Provider/Blizzard.php
@@ -25,7 +25,7 @@ class Blizzard extends OAuth2
     /**
      * {@inheritdoc}
      */
-    protected $apiBaseUrl = 'https://us.api.battle.net/';
+    protected $apiBaseUrl = 'https://us.battle.net/';
 
     /**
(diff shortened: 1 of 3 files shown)


ApacheEx:
thanks a lot 🥇

PR #1026, opened 2018-12-31
Proper Steam profile URL

bologer:
| Fixed Issues?            | Yes
| Patch: Bug Fix?          | Somewhat
| Documentation PR         | Nah

Steam does not provide any data except identity of the user as they consider this as "sensitive data". 

From the provided identity, it is possible to build profile URL and the old implementation did not work very well as identity itself was as the full profile URL and profile URL was empty. 

Now identity is the ID itself and profile URL is real profile URL.


ApacheEx:
looks like there are a lot of unnecessary changes. Let's recreate this PR only with needed fixes. Thanks

PR #1027, opened 2018-12-31
Facebook profile URL fallback

bologer:
| Patch: Bug Fix?          | Kind of
| Documentation PR         | No

Implemented fallback for Facebook to always have profile URL set on the identity user returned from the API. After sometime, Facebook stopped returning `link` from the API, so that the profile URL became empty.

### DIFF
--- a/src/Provider/Facebook.php
+++ b/src/Provider/Facebook.php
@@ -65,6 +65,11 @@ class Facebook extends OAuth2
     protected $apiDocumentation = 'https://developers.facebook.com/docs/facebook-login/overview';
 
     /**
+     * @var string Profile URL template as the fallback when no `link` returned from the API.
+     */
+    protected $profileUrlTemplate = 'https://www.facebook.com/%s';
+
+    /**
      * {@inheritdoc}
      */
     protected function initialize()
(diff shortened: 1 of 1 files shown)


src/Provider/Facebook.php
```diff
@@ -87,22 +92,29 @@ public function getUserProfile()
...
+        $userProfile->profileURL = $data->get('link');
+        $userProfile->webSiteURL = $data->get('website');
+        $userProfile->gender = $data->get('gender');
+        $userProfile->language = $data->get('locale');
         $userProfile->description = $data->get('about');
-        $userProfile->email       = $data->get('email');
+        $userProfile->email = $data->get('email');
+
+        /**
+         * Fallback for profile URL in case Facebook does not provide "pretty" link with username (if user set it).
```
ApacheEx: let's use inline style of comment

src/Provider/Facebook.php
```diff
@@ -343,15 +355,30 @@ protected function fetchUserActivity($item)
...
         return $userActivity;
     }
+
+    /**
+     * Get profile URL.
+     *
+     * @param int $identity User ID.
+     * @return string|null NULL when identity is not provided.
+     */
+    public function getProfileUrl($identity)
```
ApacheEx: or you make it as protected or you move it upper to have right order

bologer:
@ApacheEx done! :+1:

ApacheEx:
great, thank you 💯

PR #1031, opened 2019-01-10
setApiEndpoints() must be an instance of Hybridauth\Data\Collection

dusta:
| Fixed Issues?            | Update docs
| Patch: Bug Fix?          | -
| Major: Breaking Change?  | -
| Minor: New Feature?      |
| Documentation PR         | 

``` php
Fatal error: Uncaught TypeError: Argument 1 passed to Hybridauth\Adapter\AbstractAdapter::setApiEndpoints() must be an instance of Hybridauth\Data\Collection, array given

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-## [Hybridauth](https://hybridauth.github.io/) 3.0-rc8
+## [Hybridauth](https://hybridauth.github.io/) 3.0-rc9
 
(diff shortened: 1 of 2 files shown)


ApacheEx:
You are 100% right. Good catch 👍 


I think `Data\Collection` is something internal, so it's better to change a bit `setApiEndpoints` to accept array of endpoints and then wrap it to the `new Data\Collection($endpoints)`

what do you think?

dusta:
Yes, it will be more convenient. I already updated branch.

ApacheEx:
Thank you, it looks great 💯

ApacheEx:
@zorn-v I've added backward compatibility, thank you for noticing.

zorn-v:
Thanks. In future it might be deprecated, but across RCs...
BTW. When release ? 😄

ApacheEx:
@zorn-v you are right, it's ready ☺️. 

2 tasks are left: https://github.com/hybridauth/hybridauth/milestone/2
I really hope that during February, HA 3.0 will be released.

PR #1042, opened 2019-01-30
Added popup option for faceboook Authentication Window

himtweety:
we should have display=popup option enabled for the users who want to add popup support in facebook authentication

| Fixed Issues?            |
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      | enabled popup window option for facebook login
| Documentation PR         |


ApacheEx:
this won't work for sure. Because, FB supports it only for client-side application.

PR #1076, opened 2019-05-29
[LinkedIn] Upgrade to API v2

### LINKED ISSUE #1067: LinkedIn Broken with v2.14
## Bug, feature or question?

Is this a bug, a feature request or a question?
bug

Give us a short description.
The fix for the LinkedIn V2 API still hasn't made it into the v2.x branch and now LinkedIn is no longer working. This is a huge issue with anyone using this with Drupal as the Drupal module only supports v2.x of the hybridauth library.

This bug report talks about this issue, but I don't see the changes in the v2.14 release.
https://github.com/hybridauth/hybridauth/pull/1049

### Version and provider
2.14, LinkedIn

v2.x

LinkedIn

### Reproduction

ApacheEx:
| Fixed Issues?            | Fixes #1067 
| Patch: Bug Fix?          | +
| Major: Breaking Change?  | +
| Minor: New Feature?      |

This PR provides changes for API v2 of LinkedIn provider.
Mostly it's back port of #1059.

Migration guide: https://docs.microsoft.com/en-us/linkedin/consumer/integrations/self-serve/migration-faq

It's not backward compatible. All users should be upgraded to use it.

Thanks to @janadam-hbf for working on initial version 🥇

### DIFF
--- a/hybridauth/Hybrid/Providers/LinkedIn.php
+++ b/hybridauth/Hybrid/Providers/LinkedIn.php
@@ -1,9 +1,9 @@
 <?php
 
-/* !
+/*!
  * Hybridauth
  * https://hybridauth.github.io/hybridauth | https://github.com/hybridauth/hybridauth
- * (c) 2017 Hybridauth authors | https://hybridauth.github.io/license.html
+ *  (c) 2017 Hybridauth authors | https://hybridauth.github.io/license.html
  */
 
 /**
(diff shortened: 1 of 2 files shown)


janadam-hbf:
Big thanks goes also to @jakubhnilicka and @martin-klima.

natzar:
Big thanks to everyone involved in this.

One question: does anyone knows why after updating to 2.15, I just get:

**Missing provider application credentials for Provider [LinkedIn].**

But they are set as they were before, and for what I see in config.php is exactly the same.

Any clue?

ApacheEx:
@natzar make sure you use id / secret instead of key / secret

Also, if you use drupal module, there is issue but you can find workaround on drupal.org

natzar:
Many thanks! It's strange... I'm using client id, client secret. 
The "id" you mention, is client id right?

Not on drupal. Just Php, but I keep getting the same error as if keys weren't set.

natzar:
Sorry, got it.  "keys" => array("id" => xxx, "secret" => xxx) instead of "key" => , "secret" => 

Thank you very much.

PR #1094, opened 2019-06-24
Add provider for OpenStreetMap, using OAuth 1.0a, for v3

don-vip:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Minor: New Feature?      | Yes (for v3)

This PR re-adds the OpenStreetMap (OSM) provider that was added in v2 (#869) but was not ported to v3 yet.

I cherry-picked the old commits to preserve history and credit @jocelynj work instead of a copy/paste.


jocelynj:
@don-vip: you don't really need to keep my commits, but thanks anyway :)

src/Provider/OpenStreetMap.php
```diff
@@ -0,0 +1,64 @@
...
+    protected $apiDocumentation = 'https://wiki.openstreetmap.org/wiki/OAuth';
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getUserProfile()
+    {
+        $response = $this->apiRequest('user/details');
+
+        $response = new \SimpleXMLElement($response);
```
ApacheEx: looks like we now can use `$data = new Data\Collection($response);`

because it supports xml response.

ApacheEx (approved):
can you apply this remark and check if everything still works?

ApacheEx (changes requested):
thank you a lot. Would be nice to check this one thing

alexanderzatko:
sorry, I did not notice this and created a Provider for OpenStreetMap in my own PR. That one is for OAuth 2 and was tested with the 3.11.0 version of hybridauth

PR #1099, opened 2019-08-07
PSR-7 support

tobias-trozowski:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  | Yes

**Breaking Change:**
Changed signature of `Hybridauth\Adapter\AdapterInterface::authenticate` so it accepts a `Psr\Http\Message\ServerRequestInterface $request = null`.

**Description:**
Changed adapters (OpenID, OAuth1, OAuth2) accordingly to new method signature.
If a `$request !== null` is passed to `Hybridauth\Adapter\AdapterInterface::authenticate` it will return a RedirectResponse which can than be handled separately.

Allows easy integration into all PSR-15 applications.
Also solves a bug when using custom session handlers like `zendframework/zend-expressive-session` or `psr7-sessions/storageless` where sessions are persisted at the end of the execution chain (which is after `::authenticateBegin` is called where a redirect is forced).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,8 @@
         "gitter": "https://gitter.im/hybridauth/hybridauth"
     },
     "require": {
-        "php": ">=5.4.0"
+        "php": ">=5.4.0",
+        "zendframework/zend-diactoros": "^1.4||^2.0"
     },
     "require-dev": {
         "ext-curl": "*",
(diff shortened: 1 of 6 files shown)


StorytellerCZ:
Given that this is a breaking change, then this will be for HybridAuth v4.0

zorn-v:
Zend framework is `luminas` now. Maybe change dependency to actual `laminas/laminas-diactoros` ?

tobias-trozowski:
@zorn-v migrating to laminas seems to be an great idea!
if you know some lightweight alternatives for `Zend\Diactoros\Response\RedirectResponse` and `Zend\Diactoros\ServerRequestFactory` feel free to suggest something and I will have a look into it.

PR #1101, opened 2019-08-13
Update PHP Doc Blocks in Auth.php

grantlucas:
This change set updates a few Types in PHP Doc Blocks to better reflect what's possible. I was recently running a fairly large project through [Phan static analysis](https://github.com/phan/phan) and a few issues came up in the Hybrid Auth class specifically around `__construct()` and `initialize()` specifying they _only_ accept arrays. In reality they both accept both arrays or a string.

Comments have been updated to reflect this. Additionally, optional parameters that default to `null` have had their Doc Blocks updated as well.

| Patch: Bug Fix?          | Not really


StorytellerCZ:
Thank you!

PR #1103, opened 2019-08-17
Adds file storage

WinterSilence:
Stored data in temp file

| Fixed Issues?            | no
| Major: Breaking Change?  | no
| Minor: New Feature?      | yes

### DIFF
--- a/src/Storage/TempFile.php
+++ b/src/Storage/TempFile.php
@@ -1,0 +1,108 @@
+<?php
+/*!
+* Hybridauth
+* https://hybridauth.github.io | https://github.com/hybridauth/hybridauth
+*  (c) 2019 Hybridauth authors | https://hybridauth.github.io/license.html
+*/
+namespace Hybridauth\Storage;
+
+/**
+ * HybridAuth storage by temp file.
+ */
+class TempFile implements StorageInterface
+{
+    /**
+     * @var string Key prefix
+     */
+    protected $keyPrefix = '';
+
+    /**
+     * @var string Path to storage file
+     */
+    protected $path;
+
+    /**
+     * @var array Storaged data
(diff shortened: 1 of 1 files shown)


StorytellerCZ:
Can you explain a bit about this?

WinterSilence:
@StorytellerCZ temp file better than session: no output problems (cookie session),  lifetime is longer.

REJack:
There is no identifier per user like session does it in background, I guess this allows to use the saved data from the temp file even if you use a other browser/computer.
For security reasons we should use one temp file per user to seperate the prodvider infos but this would need an user identifier to set/get the right infos.

WinterSilence:
@REJack [adds `md5(get_current_user())` to filename](https://github.com/hybridauth/hybridauth/blob/bd2eefdfa753d03d23ce68c5215cbc361bcb3af7/src/Storage/TempFile.php#L36). Cookie's session have similar problem.

StorytellerCZ:
@WinterSilence Could you please also prepare a PR for the docs?

REJack:
@WinterSilence get_current_user() is not for stuff like this, it‘s gets the name of the owner of the current PHP script, it would result even to one file only. 

Cookie session doesn't have problems like this, the data’s are stored on the user's computer so only the user that logged in to a provider would have a cookie with the data and the user could only fetch his data and not from any other user.

WinterSilence:
@REJack 

pls, show example for this problem

> user could only fetch his data and not from any other use

not in all OS

REJack:
@WinterSilence I've created a demo, you can test it [here](https://hybridauth-test.rejack.space/) I logged into Facebook and you can use the "Session" on every other browser/computer.

PR #1111, opened 2019-09-09
Facebook emailVerified should not test "verified"

### LINKED ISSUE #1110: Facebook emailVerified should not test "verified"
Hello, the facebook provider tests the "verified" field of the user to return emailVerified. This is not the correct behaviour, as the verified field tests the fact that the user has verified the ownership of the phone (through SMS validation), while the email is ALWAYS verified on facebook. When Facebook returns an email, it means it has been verified.

The verified field is actually obsolete now.

Some literature : 

https://developers.facebook.com/docs/graph-api/reference/v2.0/user
> email | The User's primary email address listed on their profile. This field will not be returned if no valid email address is available.

> verified (Obsolete) | Indicates whether the account has been verified. This is distinct from the is_verified field. Someone is considered verified if they take any of the following actions:                                                                                                                                                                    
> * Register for mobile                                                                                                                                                                     
> * Confirm their account via SMS                                                                                                                                                                     
> * Enter a valid credit card

(patch follows)

bertrandgorge:
This fixes issue #1110

| Fixed Issues?            | Fixes #1110
| Patch: Bug Fix?          | Bug Fix

Uses a simple test on the email field to know if the emailVerified should be true or not.

### DIFF
--- a/src/Provider/Facebook.php
+++ b/src/Provider/Facebook.php
@@ -120,7 +120,10 @@ class Facebook extends OAuth2
 
         $userProfile->photoURL = $this->apiBaseUrl . $userProfile->identifier . '/picture?width=' . $photoSize . '&height=' . $photoSize;
 
-        $userProfile->emailVerified = $data->get('verified') == 1 ? $userProfile->email : '';
+        // Don't use $data->get('verified') here, as Facebook will only return an email if it is validated first:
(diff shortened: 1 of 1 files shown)


ApacheEx:
For sure this makes sense. Thanks a lot 🥇

PR #1130, opened 2019-12-15
Add Slack provider

cboehme:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      | Adds a provider for logging in with Slack

Adds a new provider for logging in with https://slack.com/.

In order to use it, a [Slack app](https://api.slack.com/docs/sign-in-with-slack) with the permission scopes `identity.basic`, `identity.email` and `identity.avatar` needs to be created. It is not necessary to add the app to a workspace (in fact it is not even possible if the app has only the `identity.*` scopes).

### DIFF
--- a/src/Provider/Slack.php
+++ b/src/Provider/Slack.php
@@ -1,0 +1,85 @@
+<?php
+/*!
+* Hybridauth
+* https://hybridauth.github.io | https://github.com/hybridauth/hybridauth
+*  (c) 2019 Hybridauth authors | https://hybridauth.github.io/license.html
+*/
+
+namespace Hybridauth\Provider;
+
+use Hybridauth\Adapter\OAuth2;
+use Hybridauth\Exception\UnexpectedApiResponseException;
+use Hybridauth\Data;
+use Hybridauth\User;
+
+/**
+ * Slack OAuth2 provider adapter.
+ */
+class Slack extends OAuth2
+{
+    /**
+    * {@inheritdoc}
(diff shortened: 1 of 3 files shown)


src/Provider/Slack.php
```diff
@@ -0,0 +1,85 @@
...
+ */
+class Slack extends OAuth2
+{
+    /**
+    * {@inheritdoc}
+    */
+    public $scope = 'identity.basic identity.email identity.avatar';
+
+    /**
+     {@inheritdoc}
```
ApacheEx: seems something is not good here

src/Provider/Slack.php
```diff
@@ -0,0 +1,85 @@
...
+
+        $userProfile->identifier  = $data->filter('user')->get('id');
+        $userProfile->displayName = $data->filter('user')->get('name');
+        $userProfile->email       = $data->filter('user')->get('email');
+        $userProfile->photoURL    = $this->findLargestImage($data);
+
+        return $userProfile;
+    }
+
+    private function findLargestImage(Data\Collection $data)
```
ApacheEx: some docs are needed - just for coding standards

cboehme:
Thank you for the quick response. I made the changes as requested. Additionally, I made another commit (e3e5470) which fixes a problem in the code that finds the best image. This was missed when I initially tested it.

PR #1143, opened 2020-02-28
configure() should be after setStorage()

pavkatar:
In case we send $config['tokens']['access_token'] to Facebook provider, 
./Adapter/OAuth2.php will try to save it into storage (line 251), which is not configure yet. 

```
if ($this->config->exists('tokens')) {
    $this->setAccessToken($this->config->get('tokens'));
}

public function setAccessToken($tokens = [])
{
    $this->clearStoredData();

    foreach ($tokens as $token => $value) {
        $this->storeData($token, $value);
    }

    // Re-initialize token parameters.
    $this->initialize();
}
```

| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |

### DIFF
--- a/src/Adapter/AbstractAdapter.php
+++ b/src/Adapter/AbstractAdapter.php
@@ -100,14 +100,14 @@ abstract class AbstractAdapter implements AdapterInterface
 
         $this->config = new Data\Collection($config);
 
-        $this->configure();
-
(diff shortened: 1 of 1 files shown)


ApacheEx:
you are 100% right. Interesting how we missed this.

Issue #1118, opened 2019-10-17
 Instagram Basic Graph API - new API Request Parameters


champsupertramp:
## Bug, feature or question?

Bug 

Instagram Graph API now accepts `app_id` and `app_secret` on user authentication/authorization. 

Reference: https://developers.facebook.com/docs/instagram-basic-display-api/getting-started

Instagram has a new developer page to create a new app: https://developers.facebook.com/apps/

### Version and provider

v2.16.0 
Instagram

byjujohn:
The scope has to be `user_profile,user_media` not `basic` anymore?

champsupertramp:
Hi @byjujohn  - Yes, but i'm wondering why it doesn't work with the following code? 

```
<?php
/*!
* Hybridauth
* https://hybridauth.github.io | https://github.com/hybridauth/hybridauth
*  (c) 2017 Hybridauth authors | https://hybridauth.github.io/license.html
*/

namespace Hybridauth\Provider;

use Hybridauth\Adapter\OAuth2;
use Hybridauth\Exception\UnexpectedApiResponseException;
use Hybridauth\Data;
use Hybridauth\User;

/**
 * Instagram OAuth2 provider adapter.
 */
class Instagram extends OAuth2
{
    /**
     * {@inheritdoc}
     */
    protected $scope = 'user_profile,user_media';

    /**
     * {@inheritdoc}
     */
    protected $apiBaseUrl = 'https://graph.instagram.com/';

    /**
     * {@inheritdoc}
     */
    protected $authorizeUrl = 'https://api.instagram.com/oauth/authorize/';

    /**
     * {@inheritdoc}
     */
    protected $accessTokenUrl = 'https://api.instagram.com/oauth/access_token';

    /**
     * {@inheritdoc}
     */
    protected $apiDocumentation = 'https://www.instagram.com/developer/authentication/';

    /**
     * {@inheritdoc}
     */
    protected function initialize()
    {
        parent::initialize();

        // The Instagram API requires an access_token from authenticated users
        // for each endpoint, see https://www.instagram.com/developer/endpoints.
        $accessToken = $this->getStoredData($this->accessTokenName);
        $this->apiRequestParameters[$this->accessTokenName] = $accessToken;
        $this->apiRequestParameters['app_id'] = $this->clientId;
        $this->apiRequestParameters['app_secret'] = $this->clientSecret;
        $this->AuthorizeUrlParameters['app_id'] = $this->clientId;
       
[...]

Issue #1114, opened 2019-10-02
 Instagram API breaking changes


champsupertramp:
## Bug, feature or question?

This is a question about the Instagram API roadmap.

According to the Instagram App:
```
These capabilities will be disabled immediately (previously set for July 31, 2018 or December 11, 2018 deprecation). The following will be deprecated according to the timeline we shared previously:

Public Content - all remaining capabilities to read public media on a user's behalf on December 11, 2018
Basic - to read a user’s own profile info and media in early 2020
```

I would like to know if there's a plan to migrate to Instagram Graph API? 
https://developers.facebook.com/docs/instagram-api/guides/business-discovery

Any timeline on the API migration?

### Version and provider

Latest version
Instagram

champsupertramp:
Am I too early to discuss this future breaking-changes?

champsupertramp:
I am closing this issue as I've integrated Instagram Graph API ( Facebook ).

jcambien:
Can you give us more informations ? did you built a provider by yourself ?
Thank you

champsupertramp:
Hi @jcambien  i did create the provider in the link: https://github.com/hybridauth/hybridauth/issues/1118#issuecomment-548377664

By the way, Facebook is discouraging the usage of Instagram Basic API for Login authentication.  Please the "Limitation" section in the link: https://developers.facebook.com/docs/instagram-basic-display-api#Limitations

Regards,

PR #1149, opened 2020-03-23
Added provider for Instagram via Facebook Graph API

### LINKED ISSUE #1139: Invalid scope & Invalid access_token issue with latest Instagram graph API
## Bug
This is the first error which throws invalid scope. 

`{"error_type": "OAuthException", "code": 400, "error_message": "Invalid scope: ['basic']"}`

Trying to fix this issue and got a solution for this Invalid scope.
This is your code
<pre>
class Instagram extends OAuth2
{
protected $scope = 'basic';
.........
</pre>

I changed this code to

<pre>
class Instagram extends OAuth2
{
protected $scope = 'user_profile,user_media';
.............
</pre>

These changes fixed that invalid scope issue. 

But after fixing that issue I got another issue which is Invalid access_token.

`{"meta": {"code": 400, "error_type": "OAuthAccessTokenException", "error_message": "The access_token provided is invalid."}}`

### Version and provider
I used hybridauth 3.1.1. I faced an issue with Instagram login.

### Reproduction
Those error is occurring in general. Maybe it's for the latest Instagram graph API. 

Looking forward to helping.

jcambien:
| Fixed Issues?            | Fixes #1139

Related closed issues : https://github.com/hybridauth/hybridauth/issues/1118, https://github.com/hybridauth/hybridauth/issues/1114

This introduces a new Instagram provider compatible with new API (Facebook Graph).
This work is heavily inspired by @champsupertramp example, see related issues above.

Supported features :
- Authentication
- Get user profile
- Get user medias
- Get a single user media

Our business is using it on a real application, it's working well so far.
Waiting for your feedback.


ApacheEx:
Thanks for your contribution 🥇 

a few remarks from me:
- let's fix coding standards reported by CI.
- also I think it makes sense to replace existing Instagram.php file with new one (without changing the name).
  >As of March 31, 2020, all Legacy API capabilities will be fulfilled by the Instagram Graph API and the Instagram Basic Display API.

src/Provider/Instagram.php
```diff
@@ -45,43 +45,88 @@ class Instagram extends OAuth2
     /**
      * {@inheritdoc}
      */
-    protected function initialize()
+    public function getUserProfile()
     {
-        parent::initialize();
+        $parameters = [
+            'fields' => 'id,username,account_type,media_count',
+            $this->accessTokenName => $this->getStoredData($this->accessTokenName),
```
ApacheEx: since `access_token` is required for any API calls, let's override here a method `initialize` to have something like:

```
$this->apiRequestParameters = [
    $this->accessTokenName => $this->getStoredData($this->accessTokenName),
];

File: src/Provider/Instagram.php [hunk 2]
```diff
@@ -45,43 +45,88 @@ class Instagram extends OAuth2
...
+                'id', 'caption', 'media_type', 'media_url',
+                'thumbnail_url', 'permalink', 'timestamp', 'username'
+            ];
+        }
+
+        $fields = implode(',', $fields);
 
-        $data = new Data\Collection($response);
+        $pager = $pageId !== null ? '&after=' . $pageId : '';
+        $response = $this->apiRequest('me/media?fields=' . $fields . '&limit=' . $limit . $pager);
```
ApacheEx: there is support of building query under the hood. So, just put it as 2nd arg of apiRequest.

```
$params = [
    'fields' => $fields,
    'limit' => $limit,
];
if ($pageId !== null) {
    $params['after'] = $pageId;
}

$this->apiRequest('me/media', $params);
```

p.s apply it everywhere in this file

jcambien:
There's still a error for PHP 5.5, not sure that it's related to the new code.

src/Provider/Instagram.php
```diff
@@ -45,43 +45,88 @@ class Instagram extends OAuth2
...
+     * @param int $limit Number of elements per page.
+     * @param string $pageId Current pager ID.
+     * @param array|null $fields Fields to fetch per media.
+     *
+     * @throws \Hybridauth\Exception\UnexpectedApiResponseException If API's response is not valid.
+     *
+     * @return \Hybridauth\Data\Collection Response encapsulated in a `Collection`.
      */
-    public function getUserProfile()
+    public function getUserMedia(int $limit = 12, ?string $pageId = null, array $fields = null)
```
ApacheEx: I like typehint here - but it won't work in php 5.4. so, remove int/string. and keep array only

src/Provider/Instagram.php
jcambien: I had to stay using the syntax `'fields' => implode(',', $fields)`, otherwise it don't work.

ApacheEx: yeah, true 👍 
my code is just example.

src/Provider/Instagram.php
jcambien: Fixed

src/Provider/Instagram.php
jcambien: Fixed

src/Provider/Instagram.php
jcambien: Fixed

ApacheEx:
I've made some style + doc fixes. I hope you don't mind. I will merge it as soon as CI will be ✅

ApacheEx:
Thanks!

ApacheEx:
I noticed that I changed scopes to 
image

but it doesn't work. So, I've pushed a fix to be like:
image

PR #1151, opened 2020-03-24
[VK] Add bdate field

### LINKED ISSUE #1148: How can I get a user education from Vkontakte?
This is a question.

Hello. I use v.3.

I connected to Vkontakte and got User Profile. But I don`t understood how can I get a user education?

Tell me please.

XIOLog:
| Fixed Issues?            | Fixes #1148 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |

### DIFF
--- a/src/Provider/Vkontakte.php
+++ b/src/Provider/Vkontakte.php
@@ -87,9 +87,6 @@ class Vkontakte extends OAuth2
 
         // Need to store email for later use.
         $this->storeData('email', $data->get('email'));
-        
-        // Need to store bdate for later use.
-        $this->storeData('bdate', $data->get('bdate'));
     }
 
     /**
@@ -135,9 +132,16 @@ class Vkontakte extends OAuth2
         $userProfile->displayName = $data->get('screen_name');
         $userProfile->photoURL    = $data->get('has_photo') === 1 ? $data->get($photoField) : '';
 
+        // Handle b-date.
+        if ($data->get('bdate')) {
+            $bday = explode('.', $data->get('bdate'));
+            $userProfile->birthDay = (int) $bday[0];
+            $userProfile->birthMonth = (int) $bday[1];
+            $userProfile->birthYear = (int) $bday[2];
+        }
+
         $userProfile->data = [
             'education' => $data->get('education'),
-            'bdate' => $data->get('bdate')
         ];
 
         $screen_name = static::URL . ($data->get('screen_name') ?: 'id' . $data->get('id'));


src/Provider/Vkontakte.php
```diff
@@ -87,6 +87,9 @@ protected function validateAccessTokenExchange($response)
 
         // Need to store email for later use.
         $this->storeData('email', $data->get('email'));
+        
+        // Need to store bdate for later use.
+        $this->storeData('bdate', $data->get('bdate'));
```
ApacheEx: I think this has no sense. Only email should be here as it's not regular field

src/Provider/Vkontakte.php
```diff
@@ -134,6 +137,7 @@ public function getUserProfile()
 
         $userProfile->data = [
             'education' => $data->get('education'),
+            'bdate' => $data->get('bdate')
```
ApacheEx: we have such fields like birthDay / birthMonth / birthYear. Let's use them (see https://github.com/hybridauth/hybridauth/blob/master/src/User/Profile.php#L99)

XIOLog: This fields are empty.

troubleVK3

It was a temporary solution.

ApacheEx: it's empty, because there are no code to fill them. Pls extend it to use these fields instead of $userProfile->data['bday']

XIOLog:
```php
$adapter->getUserProfile()->birthDay;
```
Right?

ApacheEx: once more - you need to fill these fields like

```php
$bday = explode('.', $data->get('bdate'));
$userProfile->birthDay = (int) $result[0];
$userProfile->birthMonth = (int) $result[1];
$userProfile->birthYear = (int) $result[2];
```


and finally, `$adapter->getUserProfile()->birthDay;` will work

XIOLog: Wow. Thank you :) I didn`t know about it.

XIOLog: I`m confused. Where should I do it?

XIOLog: I`m tried do this on $adapter, but I got an error:

Fatal error: Uncaught Error: Call to undefined method Hybridauth\Provider\Vkontakte::get()

XIOLog:
I understood. Thank you :)

PR #1155, opened 2020-03-25
[OK] Add scope "GET_EMAIL"

XIOLog:
Odnoklassniki does not provide the user's email openly. You must contact technical support to get your rights.

I contacted them and they gave me rights and wrote:

"You need to request GET_EMAIL in the scope parameter when performing OAuth authorization and fields=email when getting user information."


ApacheEx:
since it's not public available you can set scope it via config (see documentation https://hybridauth.github.io/developer-ref-user-authentication.html)

PR #1161, opened 2020-04-07
Upgrade facebook graph version to v6.0

### LINKED ISSUE #1160: Facebook Graph API v2.12 deprecation on 5th May 2020
The Facebook Graph API v2.12 which is currently used by HybridAuth v3.2.0 will be deprecated on 5th May 2020.

I've updated the code locally (in src/Provider/Facebook.php - $apiBaseUrl) to v6.0 (the latest one appearing in FB's developer portal) and was able to login with an existing user to an existing app. I've not in anyway read the docs so don't know if it will cause wider problems, but the minimal test passes :-p

### Version and provider

**v3.2.0**

**Facebook**

coling:
Fixes #1160 

Only tested to ensure I can still login on an existing application.

### DIFF
--- a/src/Provider/Facebook.php
+++ b/src/Provider/Facebook.php
@@ -47,7 +47,7 @@ class Facebook extends OAuth2
     /**
      * {@inheritdoc}
      */
-    protected $apiBaseUrl = 'https://graph.facebook.com/v2.12/';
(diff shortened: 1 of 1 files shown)


ApacheEx:
Thanks 🥇

PR #1163, opened 2020-04-24
Remove oauth1 authorization parameters from request body

collizo4sky:
I had an issue where the AWeber adapter I built was throwing a 401 error because the oauth parameters like oauth_signature were being added to the request body.

Removing that line fixed the issue.

| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |

### DIFF
--- a/src/Adapter/OAuth1.php
+++ b/src/Adapter/OAuth1.php
@@ -583,7 +583,6 @@ abstract class OAuth1 extends AbstractAdapter implements AdapterInterface
         );
 
         $uri        = $request->get_normalized_http_url();
-        $parameters = array_replace($parameters, $request->parameters);
         $headers    = array_replace($request->to_header(), (array) $headers);
 
         $response = $this->httpClient->request(


ApacheEx:
Hello,

if I understood it correctly, these oauth1 params exist in header as `Authorization`. So, we don't need to duplicate them as query params or in request body.

right?

in fact, there are only 2 provides which use Oauth1 flow: Twitter and Tumblr.
So, I think this change is safe, but would be nice if you can check if Twitter still works.

collizo4sky:
> if I understood it correctly, these oauth1 params exist in header as Authorization. So, we don't need to duplicate them as query params or in request body.

Correct.

I can confirm it works for Twitter. Also verifiable from https://developer.twitter.com/en/docs/basics/authentication/oauth-1-0a

image

Should work for Tumblr too.

ApacheEx:
Thanks! 🥇

yanilove10:
HTTP/1.1 200 OK
    Content-Type: application/json; charset=utf-8
    Content-Length: 127
    open url[1.1 200 OK     Content-Type: application/json; charset=utf-8     Content-Length: 127     ...      123{"access_token":"ACCESS_TOKEN"}](    HTTP/1.1 200 OK     Content-Type: application/json; charset=utf-8     Content-Length: 127     ...      #123 {"access_token":"ACCESS_TOKEN"}) #1319 

    {"access_token":"ACCESS_TOKEN"}

Issue #1167, opened 2020-05-01
Twitter error : code":32,"message":"Could not authenticate you."


buzzybuzza:
## Bug

When using the Twitter provider, the error
code":32,"message":"Could not authenticate you."
is returned when getUserProfile is called.

### Version and provider

Version 3.3 with Twitter provider

### Reproduction

I get that error whenever I use that provider.

buzzybuzza:
I found a fix for this. The problem is with the include_email parameter.

If the master branch, the value for this parameter is always True. Is there a reason for that?
$response = $this->apiRequest('account/verify_credentials.json?include_email=true');

In the v2 branch the parameter depends on the config
$includeEmail = isset($this->config['includeEmail']) ? (bool) $this->config['includeEmail'] : false;
		$response = $this->api->get('account/verify_credentials.json'. ($includeEmail ? '?include_email=true' : ''));

The problem is because the include_email parameter is sent in the query string.
This fix for the master branch solve the problem :
 $response = $this->apiRequest('account/verify_credentials.json', 'GET', [ 'include_email' => 'true']);

ApacheEx:
Hello, thanks for report.

I think this makes sense (+ include_email is configurable now):
```php
$response = $this->apiRequest('account/verify_credentials.json', 'GET', [
    'include_email' => $this->config->get('include_email') === false ? 'false' : 'true',
]);
```

p.s would you like to create PR?

buzzybuzza:
I created a PR.
Have a nice day!

ApacheEx:
good, merged to master. Pls check

PR #1168, opened 2020-05-01
Fix error : code":32,"message":"Could not authenticate you."

buzzybuzza:
| Fixed Issues?            | #1167
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |

Fixes the error code":32,"message":"Could not authenticate you." by sending the include_email parameter in an array instead of directly in the query string.

### DIFF
--- a/src/Provider/Twitter.php
+++ b/src/Provider/Twitter.php
@@ -81,9 +81,9 @@ class Twitter extends OAuth1
      */
     public function getUserProfile()
     {
-		$response = $this->apiRequest('account/verify_credentials.json', 'GET', [
-    		'include_email' => $this->config->get('include_email') === false ? 'false' : 'true',
-		]);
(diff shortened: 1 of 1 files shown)


src/Provider/Twitter.php
```diff
@@ -81,7 +81,7 @@ protected function getAuthorizeUrl($parameters = [])
      */
     public function getUserProfile()
     {
-        $response = $this->apiRequest('account/verify_credentials.json?include_email=true');
+        $response = $this->apiRequest('account/verify_credentials.json', 'GET', ['include_email' => $this->config->get('include_email') === false ? 'false' : 'true',]);
```
ApacheEx, line 84: let's use exactly like here:

```php
$response = $this->apiRequest('account/verify_credentials.json', 'GET', [
    'include_email' => $this->config->get('include_email') === false ? 'false' : 'true',
]);
```

PR #1169, opened 2020-05-04
Make all types in Profile nullable

BackEndTea:

All of these properties can be null, as far as i can see.  Adding the correct type hints allows static analysis tools/IDEs  to understand this.


ApacheEx:
it's good, thanks!

BackEndTea:
Thank you @ApacheEx

Issue #1170, opened 2020-05-07
Provider TwitchTV broken due to missing request header Client-ID


NeoVG:
Version: 3.3.0

**Problem:**

Apparently Twitch now needs the request header `Client-ID` additionally to `Authorization`.

If it's missing, `TwitchTV::apiRequest()` returns the following response:

`{"error":"Unauthorized","status":401,"message":"Client ID and OAuth token do not match"}.`

Here is the official statement on Twitch's developer pages regarding the change, which became active on May 1st: https://discuss.dev.twitch.tv/t/requiring-oauth-for-helix-twitch-api-endpoints/23916

**Solution:**

Just add `$this->clientId` to the appropriate request header.

I will prepare a pull request as soon as I've learned how to do it properly. :)

ApacheEx:
it's merged to master, thx. Will be released as 3.4 later.

PR #1171, opened 2020-05-07
Fix "Client ID and OAuth token do not match" with provider TwitchTV

NeoVG:
**Problem:**

Apparently Twitch now needs the request header `Client-ID` additionally to `Authorization`.

If it's missing, `TwitchTV::apiRequest()` returns the following response:

`{"error":"Unauthorized","status":401,"message":"Client ID and OAuth token do not match"}.`

Here is the official statement on Twitch's developer pages regarding the change, which became active on May 1st: https://discuss.dev.twitch.tv/t/requiring-oauth-for-helix-twitch-api-endpoints/23916

Solution:

Just add `$this->clientId` to the appropriate request header.

Issue: https://github.com/hybridauth/hybridauth/issues/1170

### DIFF
--- a/src/Provider/TwitchTV.php
+++ b/src/Provider/TwitchTV.php
@@ -45,6 +45,16 @@ class TwitchTV extends OAuth2
     /**
      * {@inheritdoc}
      */
+    protected function initialize()
+    {
+        parent::initialize();
+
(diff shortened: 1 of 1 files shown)


ApacheEx:
looks great. Thanks 🥇

Issue #1095, opened 2019-06-25
Add support for "Sign In with Apple"


smokris:
Would the Hybridauth maintainers be open to including support for the [Sign In with Apple](https://developer.apple.com/sign-in-with-apple/) service?

Additional info: https://developer.apple.com/documentation/signinwithapplejs/configuring_your_webpage_for_sign_in_with_apple

ApacheEx:
I agree that would be great.

All needed info can be found here: https://developer.apple.com/documentation/signinwithapplejs/incorporating_sign_in_with_apple_into_other_platforms

Seems, it's not a big deal to do it.

ashenwolf:
Just wondering if anyone already managed to implement Apple sign-in as hybridauth provider?

hhoweson:
Any progress on this? I would like to implement hybridauth on my site but require apple login to be available ):

ApacheEx:
this should be quite easy to implement. If no one else wants to do it I will take a look on it this weekend.

hhoweson:
That would be great! @ApacheEx

hhoweson:
@ApacheEx Any news on this? I really appreciate you taking this on and don't mean to rush you (:

hhoweson:
@ApacheEx Any news on this? I really appreciate you taking this on and don't mean to rush you (:

ApacheEx:
sorry guys, recently I've checked this but struggle with 2 things:
- from the first look we need to have possibility to encode/decode JWT. I guess would be nice to avoid this 3rd party dependency.
- to check if it works properly it requires to have Apple Developer certificate but I don't have it 😭

hhoweson:
@ApacheEx If you're able to get the rest of it working I would be happy to test it with my Apple Developer certificate

sputnik73:
@ApacheEx thank you, I will try to do some testing with our certificate, too.

hhoweson:
@ApacheEx Any updates? Let me know if there is anything I can do to help!

sputnik73:
I managed to write my own small Apple provider-class meanwhile, at least authentication is already working. 

However right now I simply use a static secret created by client_secret.rb Script from https://developer.okta.com/blog/2019/06/04/what-the-heck-is-sign-in-with-apple

Unfortunately `getUserProfile()` will not be available since Apple only delivers _email, firstname, lastname_ a single time with the first authorization request.

[...]

PR #1175, opened 2020-05-15
Add apple provider

endisha:
Hello,

Apple provider adapter.

configuration:
```
 $config = [
 'callback' => Hybridauth\HttpClient\Util::getCurrentUrl(),
  'keys'     => [ 'private_key' => '', 'id' => '', 'team_id' => '', 'key_id' => '', 'secret' => '' ],
 ];

$adapter = new Hybridauth\Provider\Apple( $config );
$adapter->authenticate();
$accessToken = $adapter->getAccessToken();
print_r($accessToken);
```

you need to create an Apple private key for Sign in with Apple service first.
create client_secret you can use https://github.com/lcobucci/jwt library to generate JWT with ES256 encryption.
```
$privateKey = file_get_contents(__DIR__.'/private_key.pem');
$clientId = '';
$teamId = '';
$keyId = '';
$alg = 'ES256';
$secret = (string) (new Builder())
						->withHeader('kid', $keyId)
						->withHeader('type', 'JWT')
						->withHeader('alg', $alg)
						->issuedBy($teamId) 
						->issuedAt(time()) 
                        ->expiresAt(time() + 3600)
                        ->permittedFor('https://appleid.apple.com') 
                        ->relatedTo($clientId)
                        ->getToken(new Sha256(), new Key($privateKey)); 
```

### DIFF
--- a/src/Provider/Apple.php
+++ b/src/Provider/Apple.php
@@ -1,0 +1,109 @@
+<?php
+/*!
+* Hybridauth
+* https://hybridauth.github.io | https://github.com/hybridauth/hybridauth
+*  (c) 2017 Hybridauth authors | https://hybridauth.github.io/license.html
+*/
+
+use Hybridauth\Adapter\OAuth2;
+
+/**
+ * Apple OAuth2 provider adapter.
+ *
+ * Example:
+ *
+ *   $config = [
+ *       'callback' => Hybridauth\HttpClient\Util::getCurrentUrl(),
+ *       'keys'     => [ 'private_key' => '', 'id' => '', 'team_id' => '', 'key_id' => '', 'secret' => '' ],
+ *   ];
+ *
+ *   $adapter = new Hybridauth\Provider\Apple( $config );
+ *
+ *   $adapter->authenticate();
+ *
+ *   $accessToken = $adapter->getAccessToken();
+ */
+class Apple extends OAuth2
+{
+    /**
+     * {@inheritdoc}
+     */
+    public $scope = 'name email';
+
+    /**
+     * {@inheritdoc}
+     */
+    protected $apiBaseUrl = 'https://appleid.apple.com/auth/';
+
+    /**
+     * {@inheritdoc}
+     */
+    protected $authorizeUrl = 'https://appleid.apple.com/auth/authorize';
+
+    /**
+     * {@inheritdoc}
+     */
+    protected $accessTokenUrl = 'https://appleid.apple.com/auth/token';
+
+    /**
+     * {@inheritdoc}
+     */
+    protected $apiDocumentation = 'https://developer.apple.com/documentation/sign_in_with_apple';
+
+    /**
+    * {@inheritdoc}
+    */
+    protected function initialize(){
+
+        parent::initialize();
+        $this->AuthorizeUrlParameters['response_mode'] = 'form_post';
+    }
+
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getAccessToken()
+    {
+        $tokenNames = [
+            'access_token',
+            'id_token',
+            'access_token_secret',
+            'token_type',
+            'refresh_token',
+            'expires_in',
+            'expires_at',
+        ];
+
+        $tokens = [];
+
+        foreach ($tokenNames as $name) {
+            if ($this->getStoredData($name)) {
+                $tokens[ $name ] = $this->getStoredData($name);
+            }
+        }
+
+        return $tokens;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function validateAccessTokenExchange($response)
+    {
+        $collection = parent::validateAccessTokenExchange($response);
+
+        $this->storeData('id_token', $collection->get('id_token'));
+
+        return $collection;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getUserProfile()
+    {
+
+    }
+
+}
--- a/src/Provider/Apple.php
+++ b/src/Provider/Apple.php
@@ -61,7 +61,29 @@ class Apple extends OAuth2
         $this->AuthorizeUrlParameters['response_mode'] = 'form_post';
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    protected function getAuthorizeUrl($parameters = [])
+    {
+        $this->AuthorizeUrlParameters = !empty($parameters)
+            ? $parameters
+            : array_replace(
+                (array) $this->AuthorizeUrlParameters,
+                (array) $this->config->get('authorize_url_parameters')
+            );
 
+        if ($this->supportRequestState) {
+            if (!isset($this->AuthorizeUrlParameters['state'])) {
+                $this->AuthorizeUrlParameters['state'] = 'HA-' . str_shuffle('ABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890');
+            }
+
+            $this->storeData('authorization_state', $this->AuthorizeUrlParameters['state']);
+        }
+
+        return $this->authorizeUrl . '?' . http_build_query($this->AuthorizeUrlParameters, '', '&', PHP_QUERY_RFC3986);
+    }
+    
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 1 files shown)


endisha:
i removed these function to verify user.
we can add JWT third party library
 ```
   /**
     * Decodes a JWT string into a PHP object
     * 
     * @param  Data\Collection  $keys
     * @param  int $key
     * @return array
     */
	protected function decodeToken($keys, $key=0){

		$parsedKeyData = $keys->filter('keys')->get($key);
	    $parsedPublicKey= \Firebase\JWT\JWK::parseKey($parsedKeyData);
	    $publicKeyDetails = openssl_pkey_get_details($parsedPublicKey);

	    if(!isset($publicKeyDetails['key'])) {
	        throw new \Exception('Invalid public key details.');
	    }

	    return \Firebase\JWT\JWT::decode($this->getStoredData('id_token'), $publicKeyDetails['key'], array($parsedKeyData->alg));
	}

    /**
     * {@inheritdoc}
     */
    {

		$response = $this->apiRequest('keys');

        $keys = new Data\Collection($response);

        if (! $keys->exists('keys')) {
            throw new UnexpectedApiResponseException('Provider API returned an unexpected response.');
        }

		$key = 0;
		$error = [];
		do{
			try {
				$payload = $this->decodeToken($keys, $key);
				$tryAgain = false;
				$error = false;
			}catch(\Exception $e) {
				$tryAgain = $key < $keys->count();
				if($e instanceof \Firebase\JWT\ExpiredException){
					$tryAgain = false;
				}
				$error = $e->getMessage();
			    $key  ;
			}
		} while($tryAgain);

		if($error){
			throw new \Exception($error);
		}
		
		
		$data = new Data\Collection($payload);

        $userProfile = new User\Profile();
        $userProfile->email = $data->get('email');

        return $userProfile;
    }
```

File: src/Provider/Apple.php [hunk 1]
```diff
@@ -61,7 +61,29 @@ protected function initialize()
...
+        if ($this->supportRequestState) {
+            if (!isset($this->AuthorizeUrlParameters['state'])) {
+                $this->AuthorizeUrlParameters['state'] = 'HA-' . str_shuffle('ABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890');
+            }
+
+            $this->storeData('authorization_state', $this->AuthorizeUrlParameters['state']);
+        }
 
+        return $this->authorizeUrl . '?' . http_build_query($this->AuthorizeUrlParameters, '', '&', PHP_QUERY_RFC3986);
+    }
```
sputnik73, line 85: Why not simply extend OAUTH2.php instead? https://github.com/sputnik73/hybridauth/commit/9909c40a8a5d2b4b7c7ec42e8b8c8baf62233e2d

sputnik73:
According to https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_js/configuring_your_webpage_for_sign_in_with_apple Apple provides a user object containing _firstname_ and _lastname_ (however on the first successfull authentication only).

This values should be used if available in `getUserProfile()`, for example by adding  `$this->storeData('id_token', $collection->get('id_token'));` in `validateAccessTokenExchange()`

Additional the identifier should be set in `getUserprofile()`;   
`$userProfile->identifier = $data->get('sub');`

sputnik73:
When testing my Apple provider class timing problems occcured. I could solve them with setting a [sleeway for JWT](https://github.com/sputnik73/hybridauth/commit/55bc5d16921e77f06de2aefe3380514436c9b0ee) I am using _firebase/php-jwt_, but it might be available for _fproject/php-jwt_, too.

sputnik73:
[As mentioned in here](https://github.com/hybridauth/hybridauth/issues/1095#issuecomment-629222091) Apple does not provide an API for `refreshAccessToken()` has to `[check if the token is  expired.](https://github.com/sputnik73/hybridauth/commit/8573f2b46eb2d1f28d182c8978df893b7b763756)`

src/Provider/Apple.php
endisha: @sputnik73
you changed main class? i just override the method to avoid change the main oauth2 class.

endisha:
> According to https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_js/configuring_your_webpage_for_sign_in_with_apple Apple provides a user object containing _firstname_ and _lastname_ (however on the first successfull authentication only).
> 
> This values should be used if available in `getUserProfile()`, for example by adding `$this->storeData('id_token', $collection->get('id_token'));` in `validateAccessTokenExchange()`
> 
> Additionally the identifier should be set in `getUserprofile()`;
> `$userProfile->identifier = $data->get('sub');`

to decoded token you must encoding JWT and needs third party library and it's a tight coupling if you add it in `getUserProfile()` directory.

src/Provider/Apple.php
sputnik73: I understand we should avoid changing main classes. However it would prevent 20 lines of copy & paste and maybe justify an additional generic setting - I am not sure what's best practice?

sputnik73:
> > According to https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_js/configuring_your_webpage_for_sign_in_with_apple Apple provides a user object containing _firstname_ and _lastname_ (however on the first successfull authentication only).
> > This values should be used if available in `getUserProfile()`, for example by adding `$this->storeData('id_token', $collection->get('id_token'));` in `validateAccessTokenExchange()`
> > Additionally the identifier should be set in `getUserprofile()`;
> > `$userProfile->identifier = $data->get('sub');`
> 
> to decoded token you must encoding JWT and needs third party library and it's a tight coupling if you add it in `getUserProfile()` directory.

It's just an additional parameter `user` (json object) returned together with `state` and `code`. Here a quick&dirty way to add it in `getUserProfile()`

```
        if (!empty($_REQUEST['user'])) {
            $objUser = json_decode($_REQUEST['user']);
            $user = new Data\Collection($objUser);

            $name = $user->get('name');
            $userProfile->firstName = $name->firstName;
            $userProfile->lastName = $name->lastName;
            $userProfile->displayName = join(' ', array($userProfile->firstName,
                $userProfile->lastName));
        }
```

sputnik73:
> to decoded token you must encoding JWT and needs third party library and it's a tight coupling if you add it in `getUserProfile()` directory.

Sorry, I only realized now that you completely removed your code from `getUserProfile()` doing the JWT encoding. I liked your solution and think it was worth the third party dependency to have `getUserProfile()` implemented.

src/Provider/Apple.php
```diff
@@ -0,0 +1,132 @@
...
+use Hybridauth\Adapter\OAuth2;
+
+/**
+ * Apple OAuth2 provider adapter.
+ *
+ * Example:
+ *
+ *   $config = [
+ *       'callback' => Hybridauth\HttpClient\Util::getCurrentUrl(),
+ *       'keys'     => [ 'private_key' => '', 'id' => '', 'team_id' => '', 'key_id' => '', 'secret' => '' ],
```
sputnik73, line 19: If you don't want to generate the secret inside the Apple provider class (which would be **much** more comfortable to use, but adds dependencies and might have a performance impact), only `secret` and `id` are required here.

sputnik73:
@endisha thank you for the professional contribution which has no impacts to the project: no dependencies and no changes on main classes. Perfect for merge.

As a user l care less about dependencies, but want to have `getUserProfile()` implemented and I don't want to care about generating the secret myself. Of course, this adds dependencies. So I am curious what others expect from this provider class.

ApacheEx:
Thank you for your contribution, I'm appreciated. 

Yes, it has no dependencies.
Yes, it is easy to merge.

Some dependencies (e.g. https://github.com/firebase/php-jwt) can be added like we did with Oauth / OpenID:
https://github.com/hybridauth/hybridauth/tree/master/src/Thirdparty

- `getUserProfile` would be nice to have. 
- seems we need to handle `generation of secret` on our side (inside Apple provider). Hybridauth enables developers to easily build social applications.

will have a look on your both PRs soon.

ApacheEx:
closed in favor of #1178

PR #1177, opened 2020-05-21
Another Apple provider

sputnik73:
The main difference of this version is that users don’t need care about secret generation, and `getUserProfile()` returns all information Apple provides. 

Additionally it handles token expiration (refresh token is not available in the Apple API). A leeway makes JWT validation more reliable.

The provider was discussed in [issue #1095 ](https://github.com/hybridauth/hybridauth/issues/1095)

### DIFF
--- a/APPLE.md
+++ b/APPLE.md
@@ -48,8 +48,7 @@ This is your Account ID at the top right of the account information (2nd line)
 
 ## Differences to other providers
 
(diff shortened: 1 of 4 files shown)


sputnik73:
Unfortunately CoderCat\JWKToPEM is not compatible to php 5.x - I will remove this library.

PR #1178, opened 2020-05-22
Another Apple provider

sputnik73:
The main difference of this version is that users don’t need care about secret generation, and `getUserProfile()` returns all information Apple provides. 

Additionally it handles token expiration (refresh token is not available in the Apple API). A leeway makes JWT validation more reliable.

The provider was discussed in [issue #1095 ](https://github.com/hybridauth/hybridauth/issues/1095)


ApacheEx:
thank you for all your hard work. I will have a look on it soon

zorn-v:
What about possibility set key content instead of key file in options ? Maybe allow both variants (with priority for `key_content`).
Handle uploading of file for provider config maybe not so simple sometimes.

sputnik73:
> What about possibility set key content instead of key file in options ? Maybe allow both variants (with priority for `key_content`).

Thx for the suggestion. Agreed that might be useful and it's not much work, I am going to implement that.

ApacheEx:
I've moved Apple.md to https://github.com/hybridauth/hybridauth.github.io (it's available here: https://hybridauth.github.io/providers/apple.html)

sputnik73:
> this one is perfect. Looks amazing for me. Yes, there are some extra composer dependencies, but I think it's fine. Thank you

Thank you for the great feedback and for merging. Any suggestions for improvement are welcome of course.

PR #1203, opened 2020-09-02
StackExchange fix and clarification

chrisgraham:
This fixes a couple of things I found testing the StackExchange provider...

First, it needs clearly documenting the 'site' parameter is needed.

Second, weirdly you need to explicitly pass the token to get the profile, even though the Bearer token header is passed.

### DIFF
--- a/src/Provider/StackExchange.php
+++ b/src/Provider/StackExchange.php
@@ -20,7 +20,7 @@ use Hybridauth\User;
  *   $config = [
  *       'callback' => Hybridauth\HttpClient\Util::getCurrentUrl(),
  *       'keys'     => [ 'id' => '', 'secret' => '' ],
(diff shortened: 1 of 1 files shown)


src/Provider/StackExchange.php
```diff
@@ -76,7 +76,7 @@ public function getUserProfile()
     {
         $site = $this->config->get('site');
 
-        $response = $this->apiRequest('me?site=' . $site);
+        $response = $this->apiRequest('me?site=' . $site, 'GET', [ 'access_token' => $this->getStoredData('access_token') ]);
```
ApacheEx, line 79: it makes sense also to move `site` to the last argument

chrisgraham: Fixed. NB: I did test this worked.

PR #1204, opened 2020-09-02
Show which URLs are failing when an endpoint returns an error

chrisgraham:
I find it helpful to make it clear what URLs are failing when you get an endpoint error.

### DIFF
--- a/src/Adapter/OAuth1.php
+++ b/src/Adapter/OAuth1.php
@@ -593,7 +593,7 @@ abstract class OAuth1 extends AbstractAdapter implements AdapterInterface
             $multipart
         );
 
-        $this->validateApiResponse('Signed API request has returned an error');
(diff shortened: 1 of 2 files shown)


src/Adapter/OAuth1.php
```diff
@@ -593,7 +593,7 @@ protected function oauthRequest($uri, $method = 'GET', $parameters = [], $header
             $multipart
         );
 
-        $this->validateApiResponse('Signed API request has returned an error');
+        $this->validateApiResponse('Signed API request to ' . $url . ' has returned an error');
```
ApacheEx, line 596: here we don't have `$url`, but rather `$uri`

ApacheEx (changes requested):
there is small remark to fix 👍

ApacheEx (approved):
Thanks!

PR #1207, opened 2020-09-02
Support photoURL as Reddit avatar

chrisgraham:
This just adds photoURL, using the avatar set in Reddit.


ApacheEx (approved):
thank you 🥇

PR #1208, opened 2020-09-03
Document where the config settings on Vkontakte come from

chrisgraham:
Vkontakte uses non-standard terminology for the OAuth settings - so comment the mappings in the provider code file.


ApacheEx (approved):
thank you

PR #1209, opened 2020-09-03
Improve MicrosoftGraph provider

### LINKED ISSUE #1025: Support Microsoft authentication via tenant endpoints
## Bug, feature or question?

feature request

The MicrosoftGraph authenticator at the moment always uses the common tenant endpoint to authenticate a user.  This is useful to authenticate users from any tenant or ones from MSA consumer users.  However, in some cases it is desired to limited authentication to users that belong to a specific tenant.  In such cases, instead of using "common" in the authorization URL, using the tenant domain would scope such authentication, helps with selecting account if multiple are signed in, and helps the application to scope auth to a single source tenant.

### Version and provider

v3.0.0-rc.8

MicrosoftGraph

### Reproduction

In my application I just use the MicrosoftGraph for authentication and authorization for scopes openid and user.read.  The scenario is really an authentication scenario rather than an authorization scenario.  I'm planning to look into getting user information via openid rather than by calling `/me` on the graph, but haven't done that yet.  So the requirements for the app is to scope users for a specific tenant for the app rather than any user to access Microsoft Graph APIs and get access to their data.

There is no really steps to reproduce since this is more of a feature request than a bug.

chrisgraham:
| Fixed Issues?            | Fixes #1025 

The MicrosoftGraph provider is rather lacking.

It couldn't get my consumer email.
It couldn't get a phone number if one is set.
And it makes a big assumption about tenant.
It also needs some docs work, as Microsoft make all this pretty confusing.

This pull request fixes all that.

### DIFF
--- a/src/Provider/MicrosoftGraph.php
+++ b/src/Provider/MicrosoftGraph.php
@@ -26,7 +26,8 @@ use Hybridauth\User;
  *   $config = [
  *       'callback' => Hybridauth\HttpClient\Util::getCurrentUrl(),
  *       'keys'     => [ 'id' => '', 'secret' => '' ],
- *       'tenant'   => 'user', // May be 'common', 'organizations' or 'consumers'
+ *       'tenant'   => 'user',
+ *         // ^ May be 'common', 'organizations' or 'consumers' or a specific tenant ID or a domain
  *   ];
  *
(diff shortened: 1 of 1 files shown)


ApacheEx (approved):
LGTM

PR #1211, opened 2020-09-03
Reimplemented Dropbox for Hybridauth v3, using Dropbox v2 API this time

chrisgraham:
Brings back Dropbox, which was in Hybridauth v2 as an additional provider, but not Hybridauth v3 yet. Upgrades to the Dropbox v2 API this time.


ApacheEx (approved):
Perfect one. Thank you 🥇

PR #1213, opened 2020-09-05
Fix code inconsistencies across providers

chrisgraham:
| Fixed Issues?            |  No

When integrating Hybridauth, and working with providers, it was a little confusing that there were discrepancies between the provider files. For example, if one provides an example, and another doesn't, it suggests that example might show something special - which is only sometimes true.

This makes providers consistent...

1) Fixing that many make $scope public for no reason, overriding the scope of the parent class (might even be some kind of error in future PHP versions?)
2) Fix inconsistent space after opening braces sometimes (encourage consistent coding standards to be followed in future)
3) Only includes examples when those examples have something unique to show (avoid wasting readers time)
4) Always provide an API documentation link, or a note if one is not available (avoids people working on a provider Googling)
5) Fix last year copyright date in the new Dropbox provider (my bad)
6) Consistent header line comment above each provider class ("<Human name> <tech> provider adapter."
7) Make examples 100% consistent with each other apart from where there is an actual implementation difference (so readers don't think differences between providers exist when they don't)

### DIFF
--- a/src/Provider/AOLOpenID.php
+++ b/src/Provider/AOLOpenID.php
@@ -18,4 +18,9 @@ class AOLOpenID extends OpenID
     * {@inheritdoc}
     */
     protected $openidIdentifier = 'http://openid.aol.com/';
+
+    /**
+     * {@inheritdoc}
+     */
+    protected $apiDocumentation = ''; // Not available
 }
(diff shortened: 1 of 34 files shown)


ApacheEx (approved):
well job. There is unrelated CI issue, but it's good to go.

PR #1215, opened 2020-09-06
Pass locale to Facebook API

### LINKED ISSUE #1097: [BUG][3.0.0][Facebook] : gender from an user with a non english language
Hello,

I have a trouble recovering the gender from Facebook. It works very well when I get an english user profile, which return male, female or null (as specified in the documentation), but in case my user is french, for exemple, the Hybridauth UserProfile will contain "homme" or "femme".

Is it a bug or does a solution exists for my problem ?

Thanks.

chrisgraham:
| Fixed Issues?            | Fixes #1097 

To get consistent gender options in English you need to tell the API your locale (as per https://stackoverflow.com/questions/35118373/facebook-graph-api-gender).
I couldn't test this as user_gender permission needs app approval, which I don't have.
But I think it'll work and it doesn't break anything.

### DIFF
--- a/src/Provider/Facebook.php
+++ b/src/Provider/Facebook.php
@@ -102,7 +102,8 @@ class Facebook extends OAuth2
             'hometown',
             'birthday',
         ];
-        $response = $this->apiRequest('me', 'GET', [ 'fields' => implode(',', $fields), 'locale' => 'en_US' ]);
+        $locale = $this->config->get('locale') ?: 'en_US'; // Note that en_US is needed for gender fields to match convention
+        $response = $this->apiRequest('me', 'GET', [ 'fields' => implode(',', $fields), 'locale' => $locale ]);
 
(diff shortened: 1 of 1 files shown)


src/Provider/Facebook.php
```diff
@@ -102,7 +102,7 @@ public function getUserProfile()
             'hometown',
             'birthday',
         ];
-        $response = $this->apiRequest('me?fields=' . implode(',', $fields));
+        $response = $this->apiRequest('me', 'GET', [ 'fields' => implode(',', $fields), 'locale' => 'en_US' ]);
```
ApacheEx, line 105: would be nice to have this setting configurable with default value `en_US`

ApacheEx (changes requested):
thanks for PR...it makes sense, but let's improve a bit

ApacheEx (approved):

PR #1217, opened 2020-09-07
Implemented Facebook token exchange mechanism

### LINKED ISSUE #1196: Facebook Long-Lived Token support
From analyzing the Hybridauth code, I can see that refresh tokens are supported implicitly, by recognizing when the oAuth flow provides them, storing them, and using them to refresh expired tokens based on looking at the token expiry timestamps.

However, from what I understand, Facebook uses a totally different mechanism, kind of a perpetual token exchange system, documented here: https://developers.facebook.com/docs/facebook-login/access-tokens/refreshing.
I believe they do this to stop a service getting an oAuth token once and then using it to spy on users in perpetuity even if the user does not use the service. (That said, it seems to me that a dodgy service could keep exchanging tokens without user interaction anyway).

Essentially you start with a short-lived token, and you exchange it for a long-lived token. And then when users are using your service, you do an exchange to get a new token with an extended future expiry timestamp but on a schedule rather than on every interaction.

This is what I can understand from reading Facebook's confusing docs anyway!

I think we need to have support for this in Hybridauth.

chrisgraham:
| Fixed Issues?            | Fixes #1196 
| Patch: Bug Fix?          | Arguable

Implements support for Facebook token exchange, which is Facebook's alternative to refresh tokens.

### DIFF
--- a/src/Provider/Facebook.php
+++ b/src/Provider/Facebook.php
@@ -99,9 +99,9 @@ class Facebook extends OAuth2
         $exchange_by_expiry_days = $this->config->get('exchange_by_expiry_days') ?? 45;
         if ($exchange_by_expiry_days !== null) {
             $projected_timestamp = time() + 60 * 60 * 24 * $exchange_by_expiry_days;
-            if (
-                ($this->hasAccessTokenExpired() === false) && // Not expired yet
-                ($this->hasAccessTokenExpired($projected_timestamp) === true)) { // But will within projection
+            if (($this->hasAccessTokenExpired() === false) && // Not expired yet
+                ($this->hasAccessTokenExpired($projected_timestamp) === true) // But will within projection
+            ) {
                 $this->exchangeAccessToken();
             }
         }
--- a/src/Provider/Facebook.php
+++ b/src/Provider/Facebook.php
@@ -96,12 +96,10 @@ class Facebook extends OAuth2
     public function apiRequest($url, $method = 'GET', $parameters = [], $headers = [], $multipart = false)
     {
         // Handle token exchange prior to the standard handler for an API request
-        $exchange_by_expiry_days = $this->config->get('exchange_by_expiry_days') ?? 45;
+        $exchange_by_expiry_days = $this->config->get('exchange_by_expiry_days') ?: 45;
         if ($exchange_by_expiry_days !== null) {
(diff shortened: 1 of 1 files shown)


src/Provider/Facebook.php
```diff
@@ -83,6 +90,55 @@ protected function initialize()
...
+    /**
+     * {@inheritdoc}
+     */
+    public function apiRequest($url, $method = 'GET', $parameters = [], $headers = [], $multipart = false)
+    {
+        // Handle token exchange prior to the standard handler for an API request
+        $exchange_by_expiry_days = $this->config->get('exchange_by_expiry_days') ?? 45;
+        if ($exchange_by_expiry_days !== null) {
+            $projected_timestamp = time() + 60 * 60 * 24 * $exchange_by_expiry_days;
+            if (($this->hasAccessTokenExpired() === false) && // Not expired yet
```
ApacheEx, line 102: let's simplify:
```php
if (!$this->hasAccessTokenExpired() && $this->hasAccessTokenExpired($projected_timestamp)) {
```

src/Provider/Facebook.php
```diff
@@ -83,6 +90,55 @@ protected function initialize()
...
+     * @throws \Hybridauth\Exception\HttpRequestFailedException
+     * @throws InvalidAccessTokenException
+     */
+    public function exchangeAccessToken()
+    {
+        $exchangeTokenParameters = [
+            'grant_type'        => 'fb_exchange_token',
+            'client_id'         => $this->clientId,
+            'client_secret'     => $this->clientSecret,
+            'fb_exchange_token' => $this->getStoredData('access_token'),
```
ApacheEx, line 126: let's say we get long-lived token, and then it will be stored as `access_token`? - next call of `exchangeAccessToken` will use long-lived token and this will not work. Since, we need to pass short-lived token.

did I miss something?

ApacheEx (changes requested):
Thanks for making it better - let's fix a few things...

src/Provider/Facebook.php
chrisgraham: Regarding your last comment, short-lived and long-lived tokens are both are access tokens, it just varies by expiry. It's not like the 'code' parameter during OAuth flow that gets exchanged to a regular tokens.

ApacheEx (approved):
now, it's good to go. Thx!

PR #1220, opened 2020-09-07
Refactor refresh token handling to fix some messy code and bugs

### LINKED ISSUE #1218: Expired/revoked/corrupt token handling
## Bug, feature or question?

This could be considered a bug, although it is arguable.

Hybridauth is over-optimistic that it will always be able to make an API request with tokens saved.

If a token is expired, it always tries to get a new one via a refresh token. Even if there is no refresh token stored (maybe a provider doesn't provide refresh tokens).

A token could also conceivably be revoked, say if a provider has a data leak and needs to revoke existing tokens.

A provider may also have changed something to make existing tokens no longer work.

In the case of the new Facebook token exchange mechanism which I just implemented, it may be the exchange would be happening too late 

I couldn't think of a better case for new exceptions.
ExpiredOauthTokenException would be thrown when Hybridauth detects its saved token is expired, before it tries to use that token.
RefusedOauthTokenException would be thrown if an API provides a response indicating a token is invalid. I'd have to research whether that is possible though, as it may not be standardized and too burdensome to implement for each provider's API.

Alternatively to ExpiredOauthTokenException, we could modify isConnected to return false if there is no refresh token and the regular token saved is expired. Actually I might even prefer that approach, because it allows automatic restart of the OAuth authentication flow. So, thoughts?

EDIT:
On reflection, I think just improving isConnected enough will be enough. Then if some code that tries to use an old authentication connection fails to call an API for some reason, it can just do a silent disconnect and the user will need to restart the login flow without being bothered.

EDIT:
On checking the code, I can see that some providers are overriding token refresh with a null-op if they don't support it. And some are re-implementing isConnected as I suggested. But it's real messy and needs a cleanup, so I'm going to make a pull request.

### Version and provider

3.4.

### Reproduction

Use an OAuth service with no refresh tokens, then try and use an API.

chrisgraham:
| Fixed Issues?            | Fixes #1218 
| Patch: Bug Fix?          | Arguably yes
| Minor: New Feature?      | Debatable

The refresh token handling code in Hybridauth has some issues, which this patch fixes.

1) At a basic level, OAuth2 is assuming every OAuth implementation supports refresh tokens, and it tries refreshing expired tokens even when it has no refresh token.
2) Some providers then override this with null-op functionality, which is inconsistent. For example, Google provider didn't override it but my testing shows it does not have refresh tokens.
3) isConnected is implemented in AbstractAdapter with hard-baked assumption that then has to be overridden away for OpenID. Messy.
4) For OAuth2 isConnected should return false if token is expired and refresh is not available
5) Telegram isConnected is also a mess. It has its own adapter API (i.e. not derived from OAuth) yet is inheriting that dodgy isConnected from AbstractAdapter. It needs its data handling rewriting so that it can store the connection info as other adapters do.
6) hasAccessTokenExpired is overriden for VK, which is good, but it has a slightly misleading code comment. Fix that.

### DIFF
--- a/src/Adapter/AbstractAdapter.php
+++ b/src/Adapter/AbstractAdapter.php
@@ -126,6 +126,11 @@ abstract class AbstractAdapter implements AdapterInterface
     /**
      * {@inheritdoc}
      */
+    abstract public function isConnected();
+
+    /**
+     * {@inheritdoc}
+     */
     public function apiRequest($url, $method = 'GET', $parameters = [], $headers = [], $multipart = false)
     {
         throw new NotImplementedException('Provider does not support this feature.');
@@ -178,11 +183,6 @@ abstract class AbstractAdapter implements AdapterInterface
     {
         throw new NotImplementedException('Provider does not support this feature.');
     }
-
-    /**
-     * {@inheritdoc}
-     */
-    public abstract function isConnected();
 
     /**
      * {@inheritdoc}
--- a/src/Adapter/AbstractAdapter.php
+++ b/src/Adapter/AbstractAdapter.php
@@ -182,6 +182,11 @@ abstract class AbstractAdapter implements AdapterInterface
     /**
      * {@inheritdoc}
      */
+    public abstract function isConnected();
+
+    /**
+     * {@inheritdoc}
+     */
     public function disconnect()
     {
         $this->clearStoredData();
--- a/src/Adapter/AbstractAdapter.php
+++ b/src/Adapter/AbstractAdapter.php
@@ -181,16 +181,6 @@ abstract class AbstractAdapter implements AdapterInterface
 
     /**
      * {@inheritdoc}
-     *
-     * Checking access_token only works for oauth1 and oauth2, openid will overwrite this method.
-     */
-    public function isConnected()
-    {
-        return (bool) $this->getStoredData('access_token');
-    }
-
-    /**
-     * {@inheritdoc}
      */
     public function disconnect()
     {
--- a/src/Adapter/OAuth2.php
+++ b/src/Adapter/OAuth2.php
@@ -333,7 +333,10 @@ abstract class OAuth2 extends AbstractAdapter implements AdapterInterface
      */
     public function isConnected()
     {
-        return (bool)$this->getStoredData('access_token') && (!$this->hasAccessTokenExpired() || $this->isRefreshTokenAvailable());
+        if ((bool)$this->getStoredData('access_token')) {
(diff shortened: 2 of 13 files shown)


src/Adapter/AbstractAdapter.php
```diff
@@ -179,16 +179,6 @@ public function setPageStatus($status, $pageId)
         throw new NotImplementedException('Provider does not support this feature.');
     }
 
-    /**
-     * {@inheritdoc}
-     *
-     * Checking access_token only works for oauth1 and oauth2, openid will overwrite this method.
-     */
-    public function isConnected()
```
ApacheEx, line 187: instead of removing this method - you need to make it as abstract. So `{@inheritdoc}` will work

src/Provider/QQ.php
```diff
@@ -61,12 +61,12 @@ protected function initialize()
...
         parent::initialize();
 
-        $this->tokenRefreshParameters = [
-            'grant_type'    => 'refresh_token',
-            'client_id'     => $this->clientId,
-            'client_secret' => $this->clientSecret,
-            'refresh_token' => $this->getStoredData('refresh_token'),
-        ];
+        if (is_array($this->tokenRefreshParameters)) {
+            $this->tokenRefreshParameters = [
```
ApacheEx, line 65: `+=` is missing

src/Adapter/OAuth2.php
```diff
@@ -323,6 +326,29 @@ public function authenticate()
         return null;
     }
 
+    /**
+     * {@inheritdoc}
+     *
+     * Checking access_token only works for oauth1 and oauth2, openid will overwrite this method.
```
ApacheEx, line 332: I think we don't need this description - since it will inheritdoc. Pls fix above

src/Provider/Patreon.php
```diff
@@ -45,10 +45,12 @@ protected function initialize()
     {
         parent::initialize();
 
-        $this->tokenRefreshParameters += [
-            'client_id' => $this->clientId,
-            'client_secret' => $this->clientSecret,
-        ];
+        if (is_array($this->tokenRefreshParameters)) {
```
ApacheEx, line 48: I've feeling we can replace `is_array($this->tokenRefreshParameters)` to `$this->isRefreshTokenAvailable()` everywhere

ApacheEx (changes requested):
Thanks. This PR has a lot of sense. Let's fix a few nitpicks, and we are good to go.

ApacheEx (approved):
I did a small phpcs fix. Thank you!

PR #1223, opened 2020-09-08
Fix incorrect exceptions being thrown by a couple of providers

chrisgraham:
| Fixed Issues?            | 

A couple of providers throw UnexpectedValueException instead of UnexpectedApiResponseException. Fixing this should not create a compat issue because UnexpectedApiResponseException is a subclass of UnexpectedValueException.


ApacheEx (approved):
👍

PR #1224, opened 2020-09-08
AuthorizationDeniedException should be used for OAuth2

chrisgraham:
| Fixed Issues?            | 
| Patch: Bug Fix?          | Arguable
| Major: Breaking Change?  | Debatable

We need to put clear focus on an OAuth flow where the user explicitly denies the login. i.e. they click something like a Cancel or a No button. Currently OAuth1 handles this, but OAuth2 just rolls it into a generic error. Use AuthorizationDeniedException for OAuth2.


ApacheEx (approved):

ApacheEx:
fully agree 👍

PR #1225, opened 2020-09-08
Fix various bad error messages and inconsistencies in exceptions

chrisgraham:
| Fixed Issues?            | 
| Major: Breaking Change?  | Probably not

Exception messages are not correct. Mainly it complains about "invalid <some kind of token>" for situations where there is actually no token passed. Change these to accurate messages.

In one case (the second line change of OAuth1.php it says access_token when it should say oauth_token. This is assuming the initial token is consistently called the oauth_token and exchanged for an access_token. It also makes it consistent with the specific exception used.


ApacheEx (approved):
LGTM 👍

PR #1234, opened 2020-09-21
Need to support nonces for domains that have CSP enabled

chrisgraham:
| Fixed Issues?            | 

Domains may have CSP enabled, in which case script tags need a nonce. Allow this via configuration.


ApacheEx (approved):
LGTM 👍

PR #1238, opened 2020-09-22
Fix session storage, due to bug in OpenID

### LINKED ISSUE #1007: OpenID getUserProfile() returns a phpIncompleteClass object
## Bug
When trying to run `getUserProfile()` with a OpenID-powered provider, such as Steam, the OpenID service will fire an exception with the following message: `Provider returned an unexpected response`. Dug deep into the cause of this, it seems like `Hybridauth/Adapter/OpenID.php`, line 267 returns a phpIncompleteClass object, which then fails the next `is_object` check:

```php
$userProfile = $this->storage->get($this->providerId . '.user');

if (! is_object($userProfile)) {
```

I don't know why and how to solve this.

### Version and provider
Steam, version 3.0 RC 7

PHP 7.1.2

### Reproduction
Set up a regular Steam provider authentication flow. Authenticate. Try to run `getUserProfile()` when already authenticated, Hybridauth returns `Provider returned an unexpected response.`.

chrisgraham:
| Fixed Issues?            | Fixes #1007

Encapsulates objects to workaround an issue with storing them.


ApacheEx:
I think we can solve it in another way. Let's just avoid storing the object:
https://github.com/hybridauth/hybridauth/blob/64ab90a33bfe8ffc414494b0cad0aa9c470f2773/src/Adapter/OpenID.php#L177

We can store an array of data instead and then recreate object based on it. What do you think?

chrisgraham:
> I think we can solve it in another way. Let's just avoid storing the object:
> 
> https://github.com/hybridauth/hybridauth/blob/64ab90a33bfe8ffc414494b0cad0aa9c470f2773/src/Adapter/OpenID.php#L177
> 
> We can store an array of profile data instead and then recreate object based on it. What do you think?

It could be done of course, but it would involve a lot more changes and would break compatibility with existing authorizations for those who don't have the bug due to initializing sessions later.

I think logically you'd change...
```
        $userProfile = $this->fetchUserProfile($openidAttributes);

        /* with openid providers we only get user profiles once, so we store it */
        $this->storage->set($this->providerId . '.user', $userProfile);
```

to:
```
        /* with openid providers we only get user profiles once, so we store it */
        $this->storage->set($this->providerId . '.user', $openidAttributes);
```

(also changes in Steam.php and StackExchangeOpenID.php)

And you'd parse it later.

That's fine. It's cleaner. But breaking compatibility bothers me.

Perhaps accept my change, and add a new issue to do it more cleanly in a future release?

ApacheEx (approved):
>Perhaps accept my change, and add a new issue to do it more cleanly in a future release?

accepted 💯

PR #1242, opened 2020-09-23
Fix PSR-2 compliance

### LINKED ISSUE #1221: Function phpdoc comment alignment (super pedantic)
This kind of thing is bugging me:
```
    /**
    * Initiate a new session
    *
    * @throws RuntimeException
    */
```
because it should be:
```
    /**
     * Initiate a new session
     *
     * @throws RuntimeException
     */
```

It's obviously a tiny thing, but it looks wrong and it is inconsistent across the code. It's good to get consistency so that contributors can then run with that.

I'm not planning to make a pull request as big search and replaces are against your contribution guidelines (for good reason).

chrisgraham:
| Fixed Issues?            | Fixes #1221
| Patch: Bug Fix?          | Kinda

Hybridauth is documented to be using PSR-2, but in many cases it is not. Fix this consistently across the whole code base, except from third-party code.

This was done using PHPStorm. I tweaked it a little bit to restrict it to PHP, and to make sure phpcs still passes.

Obviously this is a big commit.

### DIFF
--- a/examples/example_01.php
+++ b/examples/example_01.php
@@ -10,7 +10,7 @@
  *
  * Normally this step is not required as Hybridauth will attempt to start the session for you, however
  * in some cases it might be better to call session_start() at top of script to avoid cookie-based sessions
- * issues. 
+ * issues.
  *
  * See: http://php.net/manual/en/function.session-start.php#refsect1-function.session-start-notes
  *      http://stackoverflow.com/a/8028987
@@ -57,7 +57,7 @@ include 'vendor/autoload.php';
 $config = [
     'callback' => 'https://path/to/hybridauth/examples/example_01.php', // or Hybridauth\HttpClient\Util::getCurrentUrl()
 
-    'keys' => [ 'id' => 'your-app-id', 'secret' => 'your-app-secret' ], // Your Github application credentials
+    'keys' => ['id' => 'your-app-id', 'secret' => 'your-app-secret'], // Your Github application credentials
 
     /* optional : set scope
         'scope' => 'user:email', */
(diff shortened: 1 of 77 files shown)


nightw0lv:
formatting the code is not always good for finding easy something, also i notice that the code is not taking advantage of php 7+ functions

ApacheEx:
> i notice that the code is not taking advantage of php 7+ functions

yeah, it because of this:
image

nightw0lv:
> 
> 
> > i notice that the code is not taking advantage of php 7+ functions
> 
> yeah, it because of this:
> image

tell me why somebody support something that even the creator is not supporting?

PrplHaz4:
> 
> 
> > > i notice that the code is not taking advantage of php 7+ functions
> > 
> > 
> > yeah, it because of this:
> > image
> 
> tell me why somebody support something that even the creator is not supporting?
> 
> https://www.php.net/supported-versions.php

Because one of the key benefits of open source is not being beholden to vendors dictating support lifecycles?

chrisgraham:
Ok, so I do agree we should raise the minimum requirements to whatever the PHP folks are doing. Because of what is supported, and to make sure the language can continue to evolve rather than giving us all excuses for being stuck on old versions. However...

This pull request is merely trying to fix a 'bug' - i.e. the current code is not meeting the coding standard it is meant to. Recently I actually included code in a pull request that matched current code but not PSR-2, and @ApacheEx picked up on that, which I think shows the issue we have here. Further changes are out of scope.

And, people may be (and are) running older PHP versions for what they think is a good reason. For example, RedHat purport to maintain their software by backporting security-fixes, and that includes PHP. So people can run older PHP versions and still be 'secure'. (Honestly I know that's kind of BS because I've seen they don't backport all the fixes). Or, simply people haven't had the time/budget to upgrade systems. The real world is messy, unfortunately.

I've fixed a lot of bugs in the last few weeks, and what I consider important functionality holes (e.g. the Facebook token exchange mechanism). I think we should get a new release out with all that, and then if someone wants to update the code to use newer PHP features and the coding standards to say a new minimum version, I'd be for it (but it's not something I want to do personally just due to limited time I have).

Updating the minimum PHP version is good, but it should be planned and managed well.

nightw0lv:
make new branch start version 4 to support full php 7 and keep the old guys with the old tech they trust..

miled:
> make new branch start version 4 to support full php 7 and keep the old guys with the old tech they trust..

A wise man once said: talk is cheap, show me the code. But on this instance, for this topic on particular, talk seems to be costly, a waste of everyone's time. So instead of giving directives, consider helping porting the library to whatever is current and submit actual PRs. For one, you'll finally have a fancy '`contributor`' achievement tag to show it off to your frens, and then maybe a say.

ApacheEx (approved):
finally, one standard in each file. You did a great job here. Thanks a lot 💪

nightw0lv:
thanks for stopping me from sharing it with your attitude boys, not with this attitude all of you.

chrisgraham:
> thanks for stopping me from sharing it with your attitude boys, not with this attitude all of you.

I don't think anybody is against what you are asking for, just we do not think this pull request is the right place for that, and we want to handle things in an organized and polite way.

I have created a new issue for updating the PHP baseline https://github.com/hybridauth/hybridauth/issues/1246 :).

PR #1255, opened 2020-10-26
Implemented Token Exchange for Instagram

chrisgraham:
| Fixed Issues?            | 

Implemented Token Exchange for Instagram, similar to Facebook but with a number of subtle differences.

### DIFF
--- a/src/Provider/Instagram.php
+++ b/src/Provider/Instagram.php
@@ -81,6 +81,20 @@ class Instagram extends OAuth2
      */
     public function apiRequest($url, $method = 'GET', $parameters = [], $headers = [], $multipart = false)
     {
+        $this->maintainToken();
+
+        return parent::apiRequest($url, $method, $parameters, $headers, $multipart);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function maintainToken()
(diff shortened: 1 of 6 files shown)


chrisgraham:
I added another commit, as library users may need to be able to hook up Cron to regularly extend tokens in a clean way.

src/Provider/Instagram.php
```diff
@@ -55,6 +55,85 @@ protected function initialize()
...
+            $this->storeData('expires_at', $expires_at);
+        }
+
+        return $collection;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function apiRequest($url, $method = 'GET', $parameters = [], $headers = [], $multipart = false)
```
ApacheEx, line 82: Assume that `maintainToken` should be called in apiRequest - instead of overriding this method in Facebook and Instagram providers. Can we call it here? https://github.com/hybridauth/hybridauth/blob/master/src/Adapter/OAuth2.php#L710

chrisgraham:
That's sensible, done.

ApacheEx (approved):
thanks, this is nice to have improvement.

PR #1262, opened 2020-11-10
Add new provider AutoDesk

pgroot91:
Q | A
-- | --
Fixed Issues? |  
Patch: Bug Fix? | No
Major: Breaking Change? | No
Minor: New Feature? | Yes

Add new provider AutoDesk


src/Provider/AutoDesk.php
```diff
@@ -0,0 +1,91 @@
...
+    public function getUserProfile()
+    {
+		$response = $this->apiRequest('userprofile/v1/users/@me');
+
+		$collection = new Data\Collection($response);
+
+		$userProfile = new User\Profile();
+
+		$userProfile->identifier = $collection->get('userId');
+		$userProfile->displayName = $collection->get('firstName') .''. $collection->get('lastName');
```
ApacheEx, line 81: I think here you wanted to have space instead of empty string

src/Provider/AutoDesk.php
```diff
@@ -0,0 +1,91 @@
...
+    }
+	
+    /**
+     * {@inheritdoc}
+     *
+     * See: https://forge.autodesk.com/en/docs/oauth/v2/reference/http/users-@me-GET/
+     */
+    public function getUserProfile()
+    {
+		$response = $this->apiRequest('userprofile/v1/users/@me');
```
ApacheEx, line 74: I think it's wrong intend - should be 4 spaces

pls fix whole file

ApacheEx (changes requested):
thanks for your contribution. I see a few style issues which need to be fixed.

src/Provider/AutoDesk.php
pgroot91: Fixed

src/Provider/AutoDesk.php
pgroot91: Not sure why it looks like that is should have been fixed already 2 tabs. If you check view file it seems ok?

ApacheEx: looks like your file has mix with tabs/spaces.

I have converted it to be only spaces and now it's good

ApacheEx (approved):

ApacheEx:
thanks for it. This is good to go. Let's wait for CI and then I'll merge it to master

pgroot91:
Fixed

ApacheEx:
Looks good, let's release it.


Thanks!

PR #1275, opened 2020-12-14
Direct Message to Twitter

gonzalesc:
| Fixed Issues?            |

fix: Avoid arrays because it causes duplicate keys when it generates the signature to OAuth1
feature: new method to Twitter to direct message

I have created a new method to send a direct message via Twitter.
Too, I fixed a method called "get_signable_parameters" to support Twitter direct message.

Regards


ApacheEx:
Thank you for your contribution.

A few notes from my side:
- There is no plan to have `setUserDirectMessage` method just for Twitter (I've to rollback this file). It has to be on Adapter level.
- I agree with fix for OAuthRequest but we need to handle more cases. I've improved it.

ApacheEx (approved):

PR #1277, opened 2020-12-17
Patreon provider: Make getContacts return a list of patrons

chrisgraham:
| Fixed Issues?            | 

getContacts is not implemented for Patreon. Implement it as a way to get a list of patrons.
This is slightly fitting a round peg in a square hole, but I think it does make good sense to do it.


ApacheEx (approved):

ApacheEx:
nice one! Thanks

PR #1282, opened 2020-12-30
Fix to Instagram Token exchange

chrisgraham:
| Fixed Issues?            | 

I previously implemented Instagram token exchange for the project. However, there's a flaw in my implementation which this pull request fixes.

Instagram and Facebook are subtly different, and both different from standard oAuth refresh token handling.
A long-lived Facebook token lasts forever.
A long-lived Instagram token lasts ~60 days, and needs refreshing - but it turns out the refresh mechanism is like token exchange but happens with some different parameters (and totally different from oAuth refresh tokens).

FB documentation isn't so great at describing these mechanisms.

### DIFF
--- a/src/Provider/Instagram.php
+++ b/src/Provider/Instagram.php
@@ -105,14 +105,31 @@ class Instagram extends OAuth2
      */
     public function exchangeAccessToken()
     {
-        $exchangeTokenParameters = [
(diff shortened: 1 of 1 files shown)


ApacheEx:
thx))

PR #1283, opened 2021-01-13
Move apple related deps to suggest section

zorn-v:
Sometimes dependencies will cause conflicts. In my case it was `phpseclib`. I not use apple provider so I do not need those deps.
And even if I used it - phpseclib already exist and loaded in my case.
More detail about problem here https://github.com/zorn-v/nextcloud-social-login/issues/218

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         "gitter": "https://gitter.im/hybridauth/hybridauth"
     },
     "require": {
-        "php": "^5.4 || ^7.0 || ^8.0",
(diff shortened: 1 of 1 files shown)


sputnik73:
Thank you, I guess your suggestion makes sense, hopefully users will be aware of the required packages. I added some description to the how-to in https://github.com/hybridauth/hybridauth.github.io/pull/25

ApacheEx:
yeah, let's merge it and get some experience from the community. Seems to be a good suggestion for me

Thanks everyone!

PR #1300, opened 2021-06-23
Update Facebook.php

sviluppomania:


champsupertramp:
Hi @sviluppomania I'm currently having issues when retrieving the Email. The request app permission has the email and public_profile permissions but still the email is not present.

Here's the Graph API explorer:
Screen Shot 2021-07-23 at 6 53 57 PM

Here's the Permissions and Features:
Screen Shot 2021-07-23 at 6 55 10 PM


Thanks for any help that you can give.

Regards,

ApacheEx:
@sviluppomania could you tell me what is the reason of this change and how it will help us others people?

also, do you know any breaking changes or something which could affect the end user that we need to know?.

ApacheEx:
closing as this change doesn't have a proper reason

PR #1303, opened 2021-07-13
Add Keycloak Provider

Guy-Incognito:
Minor: Add Keycloak Provider.

Add Keycloak Provider provider based on OAuth2 class.

Example config:

          'Keycloak' => [
              'enabled' => true,
              'url' => 'https://your-keycloak', // depending on your setup you might need to add '/auth'
              'realm' => 'your-realm',
              'keys' => [
                  'id' => 'client-id',
                  'secret' => 'client-secret'
              ]
          ]

### DIFF
--- a/src/Provider/Keycloak.php
+++ b/src/Provider/Keycloak.php
@@ -1,0 +1,101 @@
+<?php
+/*!
+* Hybridauth
+* https://hybridauth.github.io | https://github.com/hybridauth/hybridauth
+*  (c) 2021 Hybridauth authors | https://hybridauth.github.io/license.html
+*/
+
+namespace Hybridauth\Provider;
+
+use Hybridauth\Adapter\OAuth2;
+use Hybridauth\Exception\InvalidApplicationCredentialsException;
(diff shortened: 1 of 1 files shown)


src/Provider/Keycloak.php
```diff
@@ -0,0 +1,101 @@
...
+    protected function configure()
+    {
+        parent::configure();
+
+        if (!$this->config->exists('url')) {
+            throw new InvalidApplicationCredentialsException(
+                'You must define a provider url'
+            );
+        }
+        $this->url = $this->config->get('url');
```
ApacheEx, line 60: Thank you for PR. It looks good for me. 

I was thinking about the following: 
do we really needed `$this->url` and `$this->realm` separately in the class properties, since they are available in the config and not used anywhere except `configure` method (at least for now)

I'd purpose to get rid of them.

Guy-Incognito:
@ApacheEx Thank you for your review! I pushed an according update.

ApacheEx (approved):
Thanks. Looks good now

PR #1307, opened 2021-08-28
Add PATCH request method to Guzzle to accept body json payload

CoolmanPT:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      | x

Hi there, I'm making this pull request because one of my providers are using a PATCH method request and the body params must be a json payload instead of an array. I saw that the Curl file already has the PATCH implementation but not on Guzzle.

Cheers!


ApacheEx:
Hello, thanks for help. Looks good to me

PR #1313, opened 2021-10-23
Update China WeChat Login Use WeChat Swipe to login

Seanqe:
Fix China WeChat login method Use WeChat swipe to login (Test website:https://www.seosq.com/)


cwchristerw:
Every provider should be in src/Provider directory its not good idea to move some providers to different directory. In my opinion directory names should stay in english.

Seanqe:
> 

I'm sorry, I'm just fixing it. He didn't provide a complete catalogue.

PR #1323, opened 2022-02-15
Remove unused $separator in parseBirthday

arokettu:
| Fixed Issues?            | none
| Major: Breaking Change?  | TBD
| Minor: New Feature?      | no

This is a proposal for a code cleanup. Since Parser is an internal class it should not be considered a breaking change. Feel free to discard it if it is a breaking change


ApacheEx (approved):
Makes sense, Thanks!

PR #1326, opened 2022-04-22
Fix PHP 8.1 deprecation in OpenID.php

kekalainen:
| Fixed Issues?            |  

Fixes the following deprecation message by type casting a function parameter.

```console
Deprecated: strtolower(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/html/vendor/hybridauth/hybridauth/src/Adapter/OpenID.php on line 255
```


ApacheEx:
Thanks, looks good.

PR #1329, opened 2022-05-24
Proposal: make $supportRequestState property public so it can be chan…

w3guy:
…ge after Provider instantiation.

We are using a session storage that doesn't persist on page reload and as a result would prefer to disable support for state in some adapter.

Making this public would allow us to stop creating new classes all in a bid to disable the property.


w3guy:
What do you think @ApacheEx ?

ApacheEx:
> What do you think @ApacheEx ?

Hi, probably we can add a public method instead, smth like: setSupportState

w3guy:
> > What do you think @ApacheEx ?
> 
> Hi, probably we can add a public method instead, smth like: setSupportRequestState

That would be better. Should I modify the PR to reflect that?

ApacheEx:
Yes, please 👌

ApacheEx:
also, we can achieve this by using config inside configure method.

something like:

```php
if ($this->config->exists('supportRequestState')) {
    $this->supportRequestState = $this->config->get('supportRequestState');
}

w3guy:
> also, we can achieve this by using config inside configure method (preferable).
> 
> something like:
> 
> ```
> if ($this->config->exists('supportRequestState')) {
>     $this->supportRequestState = $this->config->get('supportRequestState');
> }
> ```

Done. Please review

ApacheEx (approved):
Looks good, will be released in 3.9 soon

PR #1335, opened 2022-06-24
fix Vkontakte bdate

netsnatch:
| Fixed Issues?            |

The problem:
```
message: Undefined array key 2
file: hybridauth/hybridauth/src/Provider/Vkontakte.php
```

```
bdate
Date of Birth. Returned in the format D.M.YYYY or D.M (if year of birth is hidden). If the entire date of birth is hidden, the field is missing from the response.
```


ApacheEx:
this provider is not supported anymore

PR #1337, opened 2022-07-06
Provide support for Firebase/JWT version 6.2

### LINKED ISSUE #1331: jwt::decode() in apple provider? (with php-jwt >= 6.2.0)
FIrebase-jwt 6.2.0 has changed the vay you call jwt::decode() with regards to how you pass in the key --> https://github.com/firebase/php-jwt
https://stackoverflow.com/questions/72278051/why-is-jwtdecode-returning-status-kid-empty-unable-to-lookup-corr

from 
use Firebase\JWT\JWT;
$token = JWT::decode($jwt, $secretKey, array('HS512'));

to
use Firebase\JWT\JWT;
use Firebase\JWT\Key;
$data = JWT::decode($token, new Key($topSecret, 'HS256');

I see the apple provider https://github.com/hybridauth/hybridauth/blob/master/src/Provider/Apple.php still seems to use the old syntax... 

...have you verified it still works and they haven't broken you? 

...I am afraid of upgrading my libraries thru composer...

sputnik73:
| Fixed Issues?            | Fixes #1331

fIrebase/jwt 6.2.0 has changed the way `jwt::decode()` is called. Because this is just a suggested package in hybridauth, the version can not be restricted in composer.json. Instead this patch adds an implementation to call `jwt::decode()` required by fIrebase/jwt 6.2.0 and later and does a version check to decide which call to use.

The patch has been tested with firebase/php-jwt version 5.5.1 and with version v6.2.0 successfully.

### DIFF
--- a/src/Provider/Apple.php
+++ b/src/Provider/Apple.php
@@ -7,8 +7,8 @@
 
 namespace Hybridauth\Provider;
 
-use Hybridauth\Exception\InvalidArgumentException;
-use Hybridauth\Exception\UnexpectedApiResponseException;
+use Exception;
+use Firebase\JWT\ExpiredException;
 use Hybridauth\Exception\InvalidApplicationCredentialsException;
 use Hybridauth\Exception\UnexpectedValueException;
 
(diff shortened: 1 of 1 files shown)


ApacheEx (approved):
Looks good, appreciate your effort 👌

dani3l3:
Thanks much! When is a new release expected to be available on packagist?

ApacheEx:
I believe today 👌

PR #1339, opened 2022-08-01
Compatibility with any php version

HenryMalas:


cwchristerw:
It's not good idea to use older PHP versions, only PHP 8.0, 8.1, 8.2 are supported by The PHP Group. In my opinion its better to force users to use supported PHP version for security.
https://www.php.net/supported-versions.php

Issue #1352, opened 2022-11-19
Feature request: add Mastodon


pixeline:
## Feature

Offer to login via Mastodon.

PR #1354, opened 2022-11-24
Add Mastodon provider

RiverVanRain:
#1352

### DIFF
--- a/src/Provider/Mastodon.php
+++ b/src/Provider/Mastodon.php
@@ -10,7 +10,6 @@ use Hybridauth\User\Profile;
 
 class Mastodon extends OAuth2
 {
-
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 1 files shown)


src/Provider/Mastodon.php
```diff
@@ -0,0 +1,106 @@
...
+        $this->authorizeUrl = $url . '/oauth/authorize';
+        $this->accessTokenUrl = $url . '/oauth/token';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getUserProfile()
+    {
+		$response = $this->apiRequest('accounts/verify_credentials', 'GET', []);
```
ApacheEx, line 49: Hi, thanks for rising PR,
in general, it looks good

can we fix some alignment issue in a few places? it should be 4 spaces

ApacheEx (changes requested):

ApacheEx (approved):
I've applied formatting if you don't mind

PR #1374, opened 2023-05-23
Update Odnoklassniki.php

zharovdv:

Fix incorrect string interpolation, getMessage in Exception is a method, not a property, so we need curly braces. With out it was "Notice: Undefined property: Exception::$getMessage".


ApacheEx:
the provider is not supported. Pull requests won't be accepted.

PR #1377, opened 2023-08-02
Bugfix scope "email" Update Apple.php

markusramsak:
solves an undefined property on null error if scope is kept with "email" instead of "name email".

solves an undefined property on null error if scope is keeped with "email" instead of "name email".


ApacheEx (approved):
looks good, thx 👌

PR #1381, opened 2023-08-10
Require a state to avoid CSRF, see issue #1380 for details

### LINKED ISSUE #1380: CSRF Security Issue with state in Adapter/OAuth2.php
## Possible CSRF security issue in OAuth2

The OAuth2 Adapter (used for a lot of Providers) does not require the presence of the state even if supportRequestState is set which allows some CSRF.

### Version and provider

The problem is present in the current version v3.10.0
It affects a lot of providers as it is in the OAuth2 base class.

### Reproduction

If you intercept the returning request which has the `code` and `state` URL parameters set, strip the `state` parameter and give it to a user with a different browser (where `$this->getStoredData('authorization_state')` is null because the user has no session).
The check`$this->getStoredData('authorization_state') != $state` does not work as expected, the authorization is accepted.

I think the check should be extended to something like `(!$state || $this->getStoredData('authorization_state') != $state)` to require a `state` in this case.

The specific line is https://github.com/hybridauth/hybridauth/blob/c84f58d4533ecef3acf240c3261ca4a683fe382d/src/Adapter/OAuth2.php#L428

chschneider:
| Fixed Issues?            | Fixes #1380
| Patch: Bug Fix?          | Fixes a CSRF issue with state not being checked properly

### DIFF
--- a/src/Adapter/OAuth2.php
+++ b/src/Adapter/OAuth2.php
@@ -425,7 +425,7 @@ abstract class OAuth2 extends AbstractAdapter implements AdapterInterface
          * http://tools.ietf.org/html/rfc6749#section-4.1.1
          */
         if ($this->supportRequestState
-            && $this->getStoredData('authorization_state') != $state
+            && (!$state || $this->getStoredData('authorization_state') != $state)
         ) {
             throw new InvalidAuthorizationStateException(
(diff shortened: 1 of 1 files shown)


ApacheEx (approved):

ApacheEx:
Thanks for your contribution

Issue #1301, opened 2021-06-25
The authorization state [state=***] of this page is either invalid or has already been consumed


otengkwame:
## Bug, feature or question?
Hi, Team

I am trying to implement an authentication with LinkedIn and I keep getting this error.

**The authorization state [state=HA-*] of this page is either invalid or has already been consumed**

I have looked through the documentation to try and fix this because I thought it has already been consumed so I can logout and reconnect again.

As at now I haven't found a solution to go around this. 

### Version and provider
The provider is LinkedIn and the version is 3.7

confettidc:
Hi, Team,

I got same error too:

[Hybridauth\Exception\InvalidAuthorizationStateException] The authorization state [state=xxxxxxxx] of this page is either invalid or has already been consumed.

I used the hybridauth 3.0 few months ago for my project.

However, I got this error when use hybridauth to login my system today.

Any suggestion and idea to solve the issue?

Thanks.

KaylaLamp:
Has anyone found a solution for this?

It randomly started happening on my sites as well, when I've made no changes. I even tried to test the demos and the same error comes through.

jamminjames:
Yes, it seems random for me too. Get it all working, then BAM. It seems to having something to do with the session storage. Whenever this happens, at the same time, Hybridauth seems to be wiping out its session storage after authentication.  Can't figure out why, don't think I've changed anything that would affect it. I really wish someone who knows would help out on this forum. Seems almost abandoned.

jamminjames:
Okay, I figured out the problem in my case, maybe it will help with yours. 

Please see my issue post #1334.

kousu:
I'm seeing the same symptoms with Discord: logging in is fine for most people most of the time, but some browsers seem to get jammed and making logging in impossible.

This is the same bug: https://github.com/zorn-v/nextcloud-social-login/issues/386

ApacheEx:
The related PR was already merged and will be released as part of 3.11 soon

mirko46:
Hi, I tried solution suggest from @kousu but I have always same error.
I've apparenty solved with this workaround:
```php
--- a/src/Adapter/OAuth2.php
+++ b/Adapter/OAuth2.php
@@ -423,6 +423,7 @@ abstract class OAuth2 extends AbstractAdapter implements AdapterInterface
         if ($this->supportRequestState
             && $this->getStoredData('authorization_state') != $state
         ) {
             $this->deleteStoredData('authorization_state');
[...]

PR #1382, opened 2023-08-13
Clear invalid authorization_state during OAuth2 login

kousu:
| Fixed Issues?            | https://github.com/hybridauth/hybridauth/issues/1301, partially

There are some rare interactions where the cached OAuth2 state can get used without the login flow succeeding, like if a timeout happens or the user hits back at just the wrong moment, or other cases which I do not understand. If this happens, the browser gets jammed and cannot login, because it's stuck with an old session cookie with the invalid state.

Erase that invalid state when it is discovered.

This is part of addressing https://github.com/hybridauth/hybridauth/issues/1301. This isn't a full fix because while this defends against the app being jammed, I do not know what the real triggers are and if there is a more direct way to avoid this.

### DIFF
--- a/src/Adapter/OAuth2.php
+++ b/src/Adapter/OAuth2.php
@@ -427,6 +427,7 @@ abstract class OAuth2 extends AbstractAdapter implements AdapterInterface
         if ($this->supportRequestState
             && (!$state || $this->getStoredData('authorization_state') != $state)
         ) {
+            $this->deleteStoredData('authorization_state');
             throw new InvalidAuthorizationStateException(
                 'The authorization state [state=' . substr(htmlentities($state), 0, 100) . '] '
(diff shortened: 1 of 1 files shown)


kousu:
I've been running this patch for 6 months in production (a small server, but a real one): https://github.com/zorn-v/nextcloud-social-login/pull/398#issuecomment-1447602820. In that case, we're using NextCloud + Discord OAuth2 login. Before the patch, I'd say about 10% of the time _when people created accounts_ (i.e. logged in with Discord for the first time) this would jam, but no other time. To clear the jam they could clear their cookies, but not everyone knows how to do that so it was pretty disruptive to trying to onboard people. After the patch, I haven't had any reports of problems since.

ApacheEx (approved):

ApacheEx:
Thanks, I think it makes sense

PR #1383, opened 2023-08-15
Support php 8.2

mhdhamouday:
Prevent error "Creation of dynamic property is deprecated."

| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |


ApacheEx (approved):
Thanks, seems reasonable.

alexey-motorny-amasty:
@ApacheEx As far as can i see - you have merged this code to the master. Could you please create new version - so we can download this fix via composer

PR #1388, opened 2023-09-16
Updating the Seznam.cz provider by the documentation

pierotto:
- Adding a usage example
- Adding phone field from scope contact-phone
- Modifying of getting an email address

Currently, the data returned by the Seznam.cz API contains an `email` field, which is not in the documentation. The email address can be composed from the `username` and `domain` fields. The `account_name` field is not in the documentation and always contains a `NULL` value.

### DIFF
--- a/src/Provider/Seznam.php
+++ b/src/Provider/Seznam.php
@@ -14,6 +14,24 @@ use Hybridauth\User;
 
 /**
  * Seznam OAuth2 provider adapter.
+ *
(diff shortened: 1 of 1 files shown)


pierotto:
@ApacheEx 🙏

sspooky13:
Hello, how does it look with this pull request? Can I/we something do to help finish this pull request? Thanks

cc @ApacheEx

PR #1399, opened 2024-03-29
Unknow provider error

vardanabhanot:
| Fixed Issues?            |  No

The provider name passed to getAdapter through the authenticate function is not converted to lowercase, so in the case when the class_exists does not autoload the class, and when the comparison is made with the file name let's say GitHub.php then the condition fails to match with the Provider name(which has not been lowercased) which returns an error "Unknown Provider".


ApacheEx (approved):

ApacheEx:
thank you, makes sense

PR #1409, opened 2024-08-12
Update Facebook.php API version from 8 to current 20

elcreator:
| Patch: Bug Fix?          | Patched API to the current version

https://developers.facebook.com/docs/graph-api


elcreator:
v8 expired in 	
November 1, 2022 according to https://developers.facebook.com/docs/graph-api/changelog/versions/
If correct version is not going to be maintained then at least setter should be added (or property should be not protected but public).

elcreator:
Yes, go ahead please.

PR #1414, opened 2024-11-26
Enhance unfound provider exception message to ease troubleshooting

odain-cbd:
| Major: Breaking Change?  | Bo
| Minor: New Feature?      | Yes (to ease configuration)

Feature description:
Sometimes it is hard to start with current library. you may encounter bad configuration related to the provider you are trying to use. in that case an exception is thrown with minimal explanation "Unknown Provider"

Fix:
The message exception is enriched with adapter missing.

### DIFF
--- a/src/Hybridauth.php
+++ b/src/Hybridauth.php
@@ -134,7 +134,7 @@ class Hybridauth
                 }
             }
             if ($adapter === null) {
-                throw new InvalidArgumentException('Unknown Provider.');
+                throw new InvalidArgumentException("Unknown Provider ($adapter).");
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/Hybridauth.php
```diff
@@ -134,7 +134,7 @@ public function getAdapter($name)
                 }
             }
             if ($adapter === null) {
-                throw new InvalidArgumentException('Unknown Provider.');
+                throw new InvalidArgumentException("Unknown Provider ($adapter).");
```
ApacheEx, line 137: not sure, see if above, it checks if adapter is null, so maybe here you should log $name

ApacheEx (approved):

PR #1415, opened 2024-11-26
Refresh token fixes

odain-cbd:
| Patch: Bug Fix?          | Yes (Google/Github/MSGraph providers)
| Major: Breaking Change? | No

**Description**:
After successful authentication, access token is properly returned by below Identity Providers: Google/Github/MSGraph. But without the refresh_token needed in case of expiration. then it is not possible to renew the access token later on. 

Current PR aims at enhancing HTTP parameters sent and received to make Google/Github/MSGraph work from authentication to access token renewal.

### DIFF
--- a/src/Provider/GitHub.php
+++ b/src/Provider/GitHub.php
@@ -45,6 +45,21 @@ class GitHub extends OAuth2
     /**
      * {@inheritdoc}
      */
+    protected function initialize()
+    {
+        parent::initialize();
+
+        if ($this->isRefreshTokenAvailable()) {
+            $this->tokenRefreshParameters += [
+                'client_id' => $this->clientId,
+                'client_secret' => $this->clientSecret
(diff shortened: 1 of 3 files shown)


ApacheEx (approved):

ApacheEx:
wow, this looks like a good improvement for these providers. Thank you

andrewsauder:
The addition of the `prompt=constent` authorize url parameters in the MicrosoftGraph provider appears to be the cause of my apps relying on hybrid auth to trigger a consent prompt **every** sign in, ignoring tenant wide permissions granted to the application, and previously accepted permissions by users. Is there a reason that this url parameter was added to this bug fix in addition to the `offline_access` scope?
```php
        $this->AuthorizeUrlParameters += [
			'prompt' => 'consent',
        ];
```

PR #1418, opened 2024-12-19
Fix php84 warnings

chschneider:
| Major: Requires PHP 7.1  | This fixes PHP 8.4 deprecation warnings but needs PHP 7.1

As fixing the "Implicitly nullable parameter" deprecation warning for PHP 8.4 requires new syntax introduced PHP 7.1 I would propose to start a v4 so the PHP requirements can be pushed to PHP 7.3 as documented in the README.md.

The commits where split to reflect the different parts of the change: The warning fixes, the changes to requirements, documentation and Travis CI configuration.

The warning fixes are trivial but the decision whether this justifies v4 (plus the decision if we bump the requirements to 7.3 or only to the strictly required 7.1 due to the new syntax) is up to you.
I couldn't think of an alternative path to fix the issue but maybe there is a better approach.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,18 +8,17 @@ dist: "bionic"
 php:
   - "7.4"
   - "7.3"
-  - "7.2"
   - "8.0"
 
 jobs:
   include:
-    - name: "PHP 5.6"
+    - name: "PHP 7.3"
       dist: "xenial"
-      php: "5.6"
-    # "php": ">=5.4.0"
-    - name: "PHP 5.4"
+      php: "7.3"
+    # "php": ">=7.3.0"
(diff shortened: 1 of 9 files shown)


chschneider:
The warnings for PHP 8.4 are fixed with the other PR.

My PR also updated the requirements as I think the new version not be compatible with PHP 5 as stated in various places (e.g. autoload.php).
And as per documentation I proposed to increase the major version to be able to increase the PHP requirements.

But at the end of the day I leave it up to you how you want to handle it, I can close my PR if the issue is considered closed.

ApacheEx:
thanks, let's keep this MR, probably just rebase it with the latest changes

and I'll think how to deal with it

PR #1423, opened 2025-02-13
Fix issues with PHP 8.4 and changes to nullable vars

supriyo-biswas:
In PHP 8.4, with exceptions being converted through a set_error_handler() method, we get errors like this:

```
Hybridauth\Adapter\AbstractAdapter::__construct(): Implicitly marking parameter $httpClient as nullable is deprecated, the explicit nullable type must be used instead in /var/www/apps/mycompiler/8528a9f9/vendor/hybridauth/hybridauth/src/Adapter/AbstractAdapter.php:93
```

Since such a design is in use by many PHP frameworks and codebases, I'd like to change the occurrences of nullable variables to use `?type x = null` instead of `type x = null`.

Thank you.

### DIFF
--- a/src/Adapter/AbstractAdapter.php
+++ b/src/Adapter/AbstractAdapter.php
@@ -92,9 +92,9 @@ abstract class AbstractAdapter implements AdapterInterface
      */
     public function __construct(
         $config = [],
-        HttpClientInterface $httpClient = null,
-        StorageInterface $storage = null,
-        LoggerInterface $logger = null
+        ?HttpClientInterface $httpClient = null,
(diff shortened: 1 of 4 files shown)


ApacheEx:
Hey, thank you for your contribution. It will be available in the next version (3.12)

alexey-motorny-amasty:
Hello @ApacheEx . Could you please provide approximately release date of 3.12 version? We are waiting for compatibility with 8.4

ApacheEx:
thanks for pushing 

will release it tomorrow

PR #1439, opened 2025-09-19
fixed #1438

### LINKED ISSUE #1438: Twitch Helix API does not returns email_verified
The current Twitch implementation does not return email_verified.
It uses the helix api endpoint, after looking around found out this endpoint does not return email_verified
And could not find any doc that says having an email could be taken as a verified email.

So I have found they do support email_verified, but with the OpenID Connect endpoint [https://dev.twitch.tv/docs/authentication/getting-tokens-oidc/ ](https://dev.twitch.tv/docs/authentication/getting-tokens-oidc/ )

Now it requires an extra request to get the email and email_verified data
$oidc = $this->apiRequest('https://id.twitch.tv/oauth2/userinfo');

So, should the whole twitch api be shifted to get data through the OpenID endpoint, or just this specific email_verified be used through that endpoint? Becuase it reuquires a little change in the scope as well from `user:read:email` to `user:read:email openid`

RiverVanRain:
Twitch supports OpenID Connect

| Fixed Issues?            | `Fixes #1438`
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |

Twitch supports OpenID Connect

### DIFF
--- a/src/Provider/TwitchTV.php
+++ b/src/Provider/TwitchTV.php
@@ -1,4 +1,5 @@
 <?php
+
 /*!
 * Hybridauth
 * https://hybridauth.github.io | https://github.com/hybridauth/hybridauth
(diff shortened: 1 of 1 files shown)


ApacheEx (approved):
good to go

thanks 🙏

